PR #1, opened 2016-02-02
Refactor to move up Producer/Consumer classes

adarsh:
# Reason for Change
- The previous `MessageQueue` class took parameters for both a Producer and Consumer, which was misleading since both needed different information.
- We wanted to be able to instantiate the Consumer or Producer independently.
# Change
- Used namespacing to create a `MessageQueue::Producer` and `MessageQueue::Consumer`.
- Added namespacing for the queue listener class as well.
- Adjust tests
- Update `README.md`.
# Minor
- Get TravisCI working against the most recent few rubies.
- Add a TravisCI badge to the `README.md`.
- Bump version number.
- Add a huge warning to the readme.


README.md
```diff
@@ -31,57 +35,118 @@ To get started,
...
+These are used to
+read messages off of the queue:
 
+```ruby
+consumer = MessageQueue::Consumer.new(
+  topic: 'topic',
+  channel: 'channel'
+).connection
+
+message = consumer.size #=> 1
```
alieander: no real reason to show assignment here

README.md
```diff
@@ -31,57 +35,118 @@ To get started,
...
+message = consumer.pop
+message.body #=>'hey this is my message!'
+message.finish
+message = consumer.size #=> 0
```
alieander: same

README.md
adarsh: Good catch! Fixed in 42b0c99

README.md
```diff
@@ -31,57 +35,118 @@ To get started,
...
+MessageQueue::Listener.new(topic: topic, channel: channel).start
+```
 
-*IMPORTANT NOTE:* You must create your own `MessageProcessor` class
-for this gem to work in your application.
+This will block until
+there is a new message on the queue,
+      pop the next message
+      off of the queue
+      and send it to `MessageProcessor.new(message).start`.
```
alieander: considering it is expected of the `start` message to `finish` the message, perhaps it should go by another name. `process`?

lib/fastly_nsq/message_queue_consumer.rb
```diff
@@ -0,0 +1,35 @@
...
+    def connection
+      consumer.new(params)
+    end
+
+    private
+
+    attr_reader :channel, :topic
+
+    def consumer
+      if ENV['FAKE_QUEUE'].nil?
```
alieander: What if I set this `false` or `0`? as is want to happen programmatically

alieander: http://www.rubydoc.info/docs/rails/4.1.7/ActiveRecord/ConnectionAdapters/Column

ActiveRecord has FALSE_VALUES and TRUE_VALUES... it might be good to check for some of those somewhere or to raise when it this is not set to something unexpected.

lib/fastly_nsq/message_queue_consumer.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module MessageQueue
+  class Consumer
+    def initialize(topic:, channel:)
+      @topic = topic
+      @channel = channel
+    end
+
+    def connection
+      consumer.new(params)
```
alieander: where does params come from?

README.md
adarsh: Agree `.start` is already loaded, meaning-wise.

On the other hand `.process` feels redundant with the class name.

How about `.go`?

lib/fastly_nsq/message_queue_listener.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    def start
+      loop do
+        process_next_message
+      end
+    end
+
+    def process_next_message
+      message = consumer.pop
+      MessageProcessor.new(message).start
```
alieander: As I said in the Documentation comment, we should call out to something else here... `process`

README.md
adarsh: Replaced `.start` with `.go` in 565a35a

lib/fastly_nsq/message_queue_listener.rb
adarsh: Yep! Should be done in 565a35a

lib/fastly_nsq/message_queue_consumer.rb
adarsh: Good call. Added this in 46baf8a and dc66324.

adarsh:
@alieander Ready for re-review!

alieander:
:dango:

PR #2, opened 2016-02-03
Remove dependancy on ActiveSupport

adarsh:
# Reason for Change
- Our dependency on `ActiveSupport` collides with some older Rails apps.
# Changes
- Remove the dependency by just writing our own `cattr_accessor`.
- Bump the version.


alieander:
This is an amazing improvement.

PR #3, opened 2016-02-03
Use a more canonical way to require files

adarsh:
# Reason for Change
- When using the gem as a git remote in bundler (either with a branch or without) in William, all files were `require`d.
- Now it’s up on rubygems.org, it does not require any of the files.
# Changes
- Use the prescribed way to require all files in the `/lib` directory. While this seems aggressive to me, I guess it's standard.


alieander:
:+1:

PR #4, opened 2016-02-09
Initialize the fake queue with an empty array

adarsh:
# Reason for Change
- When installing the gem on another application, I noticed some failures related to the lack of calling `FakeMessageQueue.reset!` before each test.
- This is due to not initializing the class queue variable at the outset.
# Changes
- Initialize the queue to `[]` to start, removing the requirement to `FakeMessageQueue.reset!` to get started.
- Change the tests to only call the reset when necessary, not every time.
- Update the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,11 @@ start by using the fake queue:
 FAKE_QUEUE=true
 ```
 
+Also be sure call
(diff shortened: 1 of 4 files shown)


alieander:
lovely

PR #6, opened 2016-02-11
To add listener rake task

alieander:
This adds in a simple RakeTask that can be used to kick off a blocking listener.

**Reasons for change:**
We have already had the need for this in several places. Baking it into the gem will help us simply execution time of new usages of this gem.

**Additions:**
- `RakeTask` that can be included in external projects for simplicity
- Tests for that RakeTask
- a .ruby-version file pointing to 2.3.0 to help inform others of the environment this was developed under
- README entry describing a bit about the usage of the RakeTask
- example `Rakefile` to help guide further


examples/Rakefile
```diff
@@ -0,0 +1,37 @@
...
+require 'fastly_nsq/rake_task'
+
+##
+# Both topic and channel will need to be passed into the rake task when
+# it is called.
+#
+# ex: `rake listen_task[my_topic,my_channel]`
+# topic will be 'my_topic' and channel will be 'my_channel'
+#
+MessageQueue::RakeTask.new(:listen_task, [:topic, :channel])
```
adarsh, line 11: As discussed IRL, make sure this runs well in a Rails app because I recall having to include some kind of `:environment` flag to include the rest of the application code - I got a lot of `Undefined Method BlahClass` but maybe it's not necessary when defining tasks in this way.

examples/Rakefile
```diff
@@ -0,0 +1,37 @@
...
+# to override at time of rake call
+#
+# ex: `rake listen_task[altered_topic]`
+# topic will be 'altered_topic' and channel will be 'some_channel'
+#
+MessageQueue::RakeTask.new(:listen_task, [:topic, :channel]) do |task|
+  task.topic   = 'some_topic'
+  task.channel = 'some_channel'
+end
+
```
adarsh: Style: Extra newline

test/lib/fastly_nsq/rake_task_test.rb
```diff
@@ -0,0 +1,64 @@
+require 'test_helper'
+require 'fastly_nsq/rake_task'
+
+describe MessageQueue::RakeTask do
+  before(:each) do
+    Rake::Task['begin_listening'].clear if Rake::Task.task_defined?('begin_listening')
+    allow_any_instance_of(MessageQueue::RakeTask).to receive(:output) { nil }
```
adarsh: Consider moving this stub into the tests themselves - it's the kind of thing that might surprise me if I didn't know about it. The line above should stay in this `before` block, though, since it's setup stuff.

adarsh:
Noice. One style fix and one test question, but otherwise GTG.

adarsh:
Ping @alieander, this is ready to squash and has one outstanding question.

alieander:
Still need to look into the outstanding questions here.
I also want to put this back into Squaw and test it there to make sure it works...

Need a bit more time.

examples/Rakefile
alieander, line 11: unsure. I think I will have to test this in squaw.

PR #7, opened 2016-02-19
Fix test errors

adarsh:
# Reason for Change
- I noticed one test file didn't have the suffix `_test`.
- This means it would not have been run since MiniTest matches on that suffix.
# Changes
- Fix the filename to append `_test`.
- Fix the errors exposed once the test was actually run.


adarsh:
Ping @alieander for review.

alieander:
Looks good

PR #8, opened 2016-02-23
Terminate producer and consumer connections

adarsh:
# Reason for Change
- The consumer and producers are leaving orphaned NSQ connections because we are not closing them.
- The listener should not care about finishing the message. Right now, it knows about the behavior of a `message`, but really should only care about JSON.
# Changes
- Have the `QueueListener` close the consumer connection after processing a message.
- Add connection termination to the consumer and producer classes.
- Extract spec helpers for using the real and fake queues.
- Correct documentation to reflect the new thinking.
- Bump the minor version.
# Minor
- Send JSON-only over to the processor.
- Move the message finishing back up to the listener. The `MessageProcessor` need not know about finishing messages. This couples the behavior to another class.
- Correct bad nesting in spec `describe` blocks.

https://jira.corp.fastly.com/browse/HYD-625

### DIFF
--- a/lib/fastly_nsq/message_queue/listener.rb
+++ b/lib/fastly_nsq/message_queue/listener.rb
@@ -6,22 +6,45 @@ module MessageQueue
     end
 
     def go
+      Signal.trap('INT') do
+        shutdown
+      end
+
+      Signal.trap('TERM') do
+        shutdown
+      end
+
       loop do
-        process_next_message
+        process_one_message
       end
     end
 
     def process_next_message
-      message = consumer.pop
-      MessageProcessor.new(message).go
+      process_one_message
+      consumer.terminate
     end
 
     private
 
+    def process_one_message
+      message = consumer.pop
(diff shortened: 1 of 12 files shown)


lib/fastly_nsq/message_queue/listener.rb
```diff
@@ -13,7 +13,9 @@ def go
 
     def process_next_message
       message = consumer.pop
-      MessageProcessor.new(message).go
+      MessageProcessor.new(message.body).go
+      message.finish
+      consumer.terminate
```
alieander: This is going to fail on every loop past the first.

adarsh: Ah yes, good catch. Suggestions on how to rearchitect this class?

alieander: `INT` and `TERM` need to be trapped in the `go` loop above.

This allow us to `ensure` the closure of the `termination` of the consumer.

PR #9, opened 2016-02-23
Use RSpec

adarsh:
# Reason for Change
- `nsq-ruby` uses RSpec, so this opens up the opportunity for an upstream patch later on.
- RSpec has many advantages over Minitest including clearer failure output, spec tagging, verifying doubles, and more broad community support.  Tagging also lets us specify groups of tests and then rspec . --tag slow to just focus them.
- Since RSpec is more popular, this encourages community contributions.
# Changes
- Rename tests from `_test` to `_spec`.
- Reconfigure tests.
- `.gitignore` the `spec/example.txt` file, which tracks recent runs and failures, allowing for `rspec --only-failures` and `rspec --last-failure`.
- Display the slowest 1 test when running the suite.


pabergman:
I don't quite follow how us using minitest or rspec has any influence on the nsq-ruby gem? We just consume it here, right?

In general I'm a big fan of consistency across projects and minitest is used pretty much everewhere else.

adarsh:
> I don't quite follow how us using minitest or rspec has any influence on the nsq-ruby gem? We just consume it here, right?

We are very likely to submit this as an upstream patch to `nsq-ruby` since it doesn't do much more than wrap the functionality.

As such, that gem [is tested using RSpec](https://github.com/wistia/nsq-ruby/blob/master/spec/spec_helper.rb), and so it will facilitate proposing a patch to `nsq-ruby` if we also use RSpec here.

alieander:
Minitest consistency is fine but in this context I think Rspec is going to be faster for us to work in and simpler in the long term if there is some merging concerns with the ruby-nsq gem.

Additionally I am concerned that the consistency argument sometimes holds us up from using effective tools. Mocha is a great example of something that no longer maintained and is yet what we continue to use for consistency.

alieander:
:+1:

PR #10, opened 2016-02-26
Implement termination of the fake connections

adarsh:
# Reason for Change
- The real `Nsq::Producer/Consumer` have a `.terminate` method which closes the connection.
- The Fake lacks this, so we see some errors.
# Changes
- Implement terminate for fake producer and consumer
- Test them using RSpec's verifying instance double. If the method does not exist, RSpec will error.


alieander:
<3

PR #12, opened 2016-03-05
Raise custom exeception for empty fake queue

adarsh:
# Reason for Change
- The fake queue will return an unhelpful exception when trying to `.pop` messages off of an empty queue.
- This is because the real queue will block until there is a message, but our fake is just an array, hence `nil`.
- That error is `NoMethodError: undefined method 'body' for nil:NilClass`.
- We should be more considerate to users of this library and explain what this means.
# Changes
- Add a custom `EmptyFakeQueueError` class with better message about what is actually going on.
- Raise this error in the case described above.
- Reraise any other errors to allow normal exception stuff to happen.
# Minor
- Replace `minitest` as a the placeholder testing channel topic name.

### DIFF
--- a/lib/fastly_nsq/message_queue/listener.rb
+++ b/lib/fastly_nsq/message_queue/listener.rb
@@ -30,6 +30,12 @@ module MessageQueue
       message = consumer.pop
       MessageProcessor.new(message.body).go
       message.finish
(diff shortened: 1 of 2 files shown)


alieander:
seems fine.

PR #13, opened 2016-03-09
Have the fake consumer block on an empty queue

adarsh:
# Reason for Change
- When using `FAKE_QUEUE=true` in the dev VM, William's listener rake task fails horribly because the fake queue does not block on an empty queue.
# Changes
- Block on the empty queue when in fake mode.
## Minor
- Update the language in a spec's comments for clarity.
# Testing
- Tested this gem branch in William and all worked nicely.

https://jira.corp.fastly.com/browse/HYD-686

### DIFF
--- a/spec/lib/fastly_nsq/fake_message_queue_spec.rb
+++ b/spec/lib/fastly_nsq/fake_message_queue_spec.rb
@@ -2,14 +2,17 @@ require 'spec_helper'
 
 RSpec.describe FakeMessageQueue do
   describe '@@queue' do
-    it 'is initalized as an empty array' do
-      expect(FakeMessageQueue.queue).to eq []
+    it 'is initalized as an empty Ruby Queue' do
+      queue = FakeMessageQueue.queue
+
+      expect(queue).to be_a Thread::Queue
+      expect(queue).to be_empty
     end
   end
 
   describe '.reset!' do
     it 'resets the fake message queue' do
-      FakeMessageQueue.queue = ['hello']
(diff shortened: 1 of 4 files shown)


spec/lib/fastly_nsq/fake_message_queue_spec.rb
```diff
@@ -92,20 +92,42 @@
...
+        )
+        popped_message = consumer.pop
+
+        expect(popped_message).to eq message
+      end
+    end
 
-      expect(popped_message). to eq message
+    context 'when there no message on the queue' do
+      it 'blocks for longer than a half second' do
```
alieander: Why longer than half a second?

Shouldn't this be longer than SECONDS_BETWEEN_QUEUE_CHECKS?

alieander:
This might be pointing to how an Array may not be the correct data structure to back our FakeQueue. A Ruby Queue already supports block `pop` operations. It could come with some added difficultly due to it being purpose built to provide "a way to synchronize communication between threads" but it might be useful here as well... Just a thought.

spec/lib/fastly_nsq/fake_message_queue_spec.rb
adarsh: Yes, although now irrelevant since everything is a `Queue` in 51e3e3d.

adarsh:
@alieander totally agree. Now everything is a queue as in 51e3e3d.

alieander:
This is wonderful! <3

PR #14, opened 2016-03-09
Bump to 0.2.2

adarsh:
- Recent changes include better queue handling with the fake case. Makes development much easier.


alieander:
:dango: :+1:

PR #16, opened 2016-03-10
Allow a logger to be defined on underlying nsq gem

alieander:
Here we allow for a passthru of `.logger=` and `.logger` to the wistia/ruby-nsq gem so that we can override the default logger of `Logger.new(nil)`. This should allow us to setup the nsq gem logger to be something like `Rails.logger` or some other logger as is needed.

Changes:
- Add two methods on `MessageQueue` (`.logger=` and `.logger`)
- Support those methods on `FakeMessageQueue`
- Tests for these methods

### DIFF
--- a/lib/fastly_nsq/message_queue.rb
+++ b/lib/fastly_nsq/message_queue.rb
@@ -8,4 +8,16 @@ require_relative 'message_queue/strategy'
 module MessageQueue
   FALSY_VALUES = [false, 0, '0', 'false', 'FALSE', 'off', 'OFF', nil]
   TRUTHY_VALUES = [true, 1, '1', 'true', 'TRUE', 'on', 'ON']
+
+  def self.logger=(logger)
+    strategy.logger = logger
+  end
+
+  def self.logger
+    strategy.logger
+  end
+
+  def self.strategy
+    Strategy.for_queue
+  end
 end
--- a/spec/lib/fastly_nsq/fake_message_queue_spec.rb
+++ b/spec/lib/fastly_nsq/fake_message_queue_spec.rb
@@ -10,6 +10,23 @@ RSpec.describe FakeMessageQueue do
     end
   end
 
+  describe '@@logger' do
+    after do
+      FakeMessageQueue.logger = Logger.new(nil)
+    end
+
+    it 'is initalized as an empty Ruby Logger' do
+      expect(FakeMessageQueue.logger).to be_a Logger
+    end
+
+    it 'can be set and retrieved' do
+      logger = double('some logger')
+      FakeMessageQueue.logger = logger
+
+      expect(FakeMessageQueue.logger).to eq logger
+    end
+  end
+
   describe '.reset!' do
     it 'resets the fake message queue' do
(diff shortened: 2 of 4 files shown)


spec/lib/fastly_nsq/fake_message_queue_spec.rb
```diff
@@ -10,6 +10,23 @@
     end
   end
 
+  describe '@@logger' do
+    after do
+      FakeMessageQueue.logger = Logger.new(nil)
```
adarsh, line 15: Crazy idea: What about a `FakeMessageQueue.reset_logger!` method which does this?

adarsh, line 15: Actually, we shouldn't do this very often so don't worry about it.

spec/lib/fastly_nsq/fake_message_queue_spec.rb
```diff
@@ -10,6 +10,23 @@
...
+    end
+
+    it 'is initalized as an empty Ruby Logger' do
+      expect(FakeMessageQueue.logger).to be_a Logger
+    end
+
+    it 'is settable via logger=' do
```
adarsh: Maybe also add "... and is retrievable"? We are testing both `.logger=` and `.logger`.

spec/lib/fastly_nsq/message_queue_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+require 'fastly_nsq'
```
adarsh: Is this a necessary requirement?

lib/fastly_nsq/message_queue.rb
```diff
@@ -8,4 +8,16 @@
...
   TRUTHY_VALUES = [true, 1, '1', 'true', 'TRUE', 'on', 'ON']
+
+  ##
+  # Passthru to wistia/nsq gem logger= method
+  def self.logger=(logger)
+    Strategy.for_queue.logger = logger
+  end
+
+  ##
+  # Passthru to wistia/nsq gem logger method
```
adarsh: This passes through to both the real and fake logger methods now, so consider removing all of these comments.

code_comments

lib/fastly_nsq/message_queue.rb
```diff
@@ -8,4 +8,16 @@
 module MessageQueue
   FALSY_VALUES = [false, 0, '0', 'false', 'FALSE', 'off', 'OFF', nil]
   TRUTHY_VALUES = [true, 1, '1', 'true', 'TRUE', 'on', 'ON']
+
+  ##
+  # Passthru to wistia/nsq gem logger= method
+  def self.logger=(logger)
+    Strategy.for_queue.logger = logger
```
adarsh: Maybe extract a method like

``` ruby
def self.logger=(logger)
  strategy.logger = logger
end

def self.strategy
  Strategy.for_queue
end
```

adarsh:
Minor stuff, but LGTM.

:sheep: it!

adarsh:
:+1:

PR #17, opened 2016-03-10
Add RSpec fake_queue: true|false flags

adarsh:
# Reason for Change
- Our tests increasingly depend on setting the queue as fake or real.
- We had these `ENV` helpers already, adding these tags makes their use easier.
# Changes
- Add RSpec around configs to use the flags `fake_queue: true` and `fake_queue: false` in specs.
- Update existing tests to use these flags.


alieander:
:+1:

PR #21, opened 2016-03-11
Revert use of a Ruby Queue in the fake

adarsh:
# Reason for Change
- The conversion to use a Ruby `Queue` in c6603df was ill-considered.
- Specifically, a `Queue` can push items in on the main process but needs to be in a thread to `.pop`.
- Really, an Array is fine and does the job with no troubles, so this reverts that change.
# List of Changes
- Describe your changes
- As a concise list of items
# Changes
- This reverts commit c6603df.
# Minor
- Use a shorter constant name and use it in the tests.
- Have the test take longer than one cycle of checking for the fake queue.
- Use the `fake_queue` flag in a spec which got missed in the conversion.
# Checklist
# Recommended Reviewers

@alieander, @standingwave

### DIFF
--- a/lib/fastly_nsq/fake_message_queue.rb
+++ b/lib/fastly_nsq/fake_message_queue.rb
@@ -1,5 +1,6 @@
 module FakeMessageQueue
   @@logger = Logger.new(nil)
+  @@queue = []
 
   def self.queue
     @@queue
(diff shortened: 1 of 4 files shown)


alieander:
I Should do.

:cry: for Ruby Queues

PR #22, opened 2016-03-17
Listen to multiple queues at once

adarsh:
# Reason for Change
- As we've set up the topic listening machinery in the `fastly_nsq` gem, it is really geared toward listening to and processing one topic's messages.
- However, in some apps, we will shortly be processing messages from two topics simultaneously, so our system needs to handle > 1 queue topics simultaneous.
# Changes
- Don't pass topics into the Raketask, specify in `MessageProcessor.topics`.
- Remove the required `event_type` key from message bodies, in the examples and the documentation.
- Have the processor accept a topic as well as a message body.
- Rework the rake task to spawn a new thread for each topic.
- Update the example Rakefile to reflect the changes and documentation.
## Minor
- Bump version to 0.3.0.
- Extract some methods in the `RakeTask`.
- Style: Avoid trailing conditionals.
- Style: Avoid unused tempvars.
- Don't include both `fastly_nsq` and `fastly_nsq/rake_task` - no need.
- Profile the slowest example on full suite RSpec.
- Move the setting of the `MessageProcessor` class to a before block.
- Clear all rake tasks before each test; use context blocks.
- Prevent output during the tests and do it in the before block.
- Correct mislabeled describe block.
- Spelling error in pull request template.
## Risks
- This represents a breaking change to the APIs for the rake task and the MessageProcessor. All apps must be updated with this version bump to accomodate the change.
- The Rake task now uses threads, one for each topic monitored.

https://jira.corp.fastly.com/browse/HYD-735
# Checklist

# Recommended Reviewers

@alieander, @standingwave

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,6 @@ write messages onto the queue:
 
 ```ruby
 message_data = {
-  "event_type" => "heartbeat",
   "data" => {
     "key" => "value"
   }
@@ -65,7 +64,6 @@ producer = MessageQueue::Producer.new(
 
 producer.write(message_data.to_json)
 ```
-
 The mock/real strategy used
 can be switched
 by adding an environment variable
@@ -94,7 +92,7 @@ consumer = MessageQueue::Consumer.new(
 
 consumer.size #=> 1
 message = consumer.pop
-message.body #=> "{ 'event_type': 'heartbeat','data': { 'key': 'value' } }"
+message.body #=> "{ 'data': { 'key': 'value' } }"
 message.finish
 consumer.size #=> 0
 consumer.terminate
@@ -119,7 +117,7 @@ MessageQueue::Listener.new(topic: topic, channel: channel).process_next_message
 This will pop the next message
 off of the queue
 and send the JSON text body
-to `MessageProcessor.new(message_body).go`.
+to `MessageProcessor.new(message_body: message_body, topic: topic).go`.
 
 To initiate a blocking loop to process messages continuously:
 
@@ -141,15 +139,23 @@ there is a new message on the queue,
 To help facilitate running the `MessageQueue::Listener` in a blocking fashion
 outside your application, a simple `RakeTask` is provided.
 
-This can be added into your `Rakefile` in one of two ways:
+NOTE: The rake task expects a
+`MessageProcessor.topics` method,
+which must return an array of strings
+defining the topics to which
+we would like to listen and process messages.
+
+The task will listen
+to all specified topics,
+each in a separate thread.
+
+This task can be added into your `Rakefile` in one of two ways:
 
 Using a block:
 ```ruby
-require 'fastly_nsq'
 require 'fastly_nsq/rake_task'
 
 MessageQueue::RakeTask.new(:listen_task) do |task|
-  task.topic   = 'some_topic'
   task.channel = 'some_channel'
 end
 
@@ -159,13 +165,12 @@ end
 
 or using passed in values:
 ```ruby
-require 'fastly_nsq'
 require 'fastly_nsq/rake_task'
 
-MessageQueue::RakeTask.new(:listen_task, [:topic, :channel])
+MessageQueue::RakeTask.new(:listen_task, [:channel])
 
 # usage:
-`rake listen_task['my_topic','my_channel']`
+`rake listen_task['my_channel']`
 ```
 
 Both methods can be used at the same time with the passed in values taking
--- a/examples/Rakefile
+++ b/examples/Rakefile
@@ -1,41 +1,88 @@
-require 'fastly_nsq'
 require 'fastly_nsq/rake_task'
 
-##
-# Both topic and channel will need to be passed into the rake task when
-# it is called.
-#
-# ex: `rake listen_task[my_topic,my_channel]`
-# topic will be 'my_topic' and channel will be 'my_channel'
-#
-MessageQueue::RakeTask.new(:listen_task, [:topic, :channel])
+#-------------------------------------------------------------------------------
+# You are required to to set the topics in MessageProcessor.topics as an array:
 
-##
-# In Rails, you can include the environment like so:
-#
-MessageQueue::RakeTask.new(:listen_task, [:topic, :channel] => :environment)
+class MessageProcessor
+  def self.topics
+    ['customer_created', 'customer_deleted']
+  end
 
-##
-# Topic and channel can also be preset in the task by passing a block and
-# assigning value like so. This is useful if you want a single purpose
-# listener task.
+  # ...
+end
+
+# A channel will need to be passed in
+# to the Rake task
+# when it is called.
 #
-# ex: `rake listen_task`
-# topic will be 'some_topic' and channel will be 'some_channel'
-#
+# The task looks like:
+
+desc 'Listen to the messaging queue with the given channel.'
+
+MessageQueue::RakeTask.new(:listen_task, [:channel])
+
+# Call `rake listen_task[my_channel]`.
+# The topics will be ['customer_created', 'customer_deleted']
+# and channel will be 'my_channel'.
+
+#-------------------------------------------------------------------------------
+# In Rails, you can include the application environment. The task looks like:
+
+desc 'Listen to the messaging queue with the given channel.'
+
+MessageQueue::RakeTask.new(:listen_task, [:channel] => :environment)
+
+# Call `rake listen_task[your_channel]`.
+# The topics will be ['customer_created', 'customer_deleted']
+# and channel will be 'your_channel'.
+
+#-------------------------------------------------------------------------------
+# The channel can also be preset in the task
+# by passing a block and assigning the value. The task looks like:
+
+desc 'Listen to the messaging queue with the given channel.'
+
 MessageQueue::RakeTask.new(:listen_task) do |task|
-  task.topic   = 'some_topic'
   task.channel = 'some_channel'
 end
 
-##
-# Both forms can be combine to provide defaults of a sort with the ability
-# to override at time of rake call
-#
-# ex: `rake listen_task[altered_topic]`
-# topic will be 'altered_topic' and channel will be 'some_channel'
-#
-MessageQueue::RakeTask.new(:listen_task, [:topic, :channel]) do |task|
-  task.topic   = 'some_topic'
-  task.channel = 'some_channel'
+# Call `rake listen_task`.
+# The topics will be ['customer_created', 'customer_deleted']
+# and channel will be 'some_channel'.
+
+#-------------------------------------------------------------------------------
+# Do the same thing and include the environment:
+
+desc 'Listen to the messaging queue with the given channel.'
+
+MessageQueue::RakeTask.new(:listen_task, [] => :environment) do |task|
+  task.channel = 'her_channel'
 end
+
+# Call `rake listen_task`.
+# The topics will be ['customer_created', 'customer_deleted']
+# and channel will be 'her_channel'.
+
+#-------------------------------------------------------------------------------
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -52,7 +52,6 @@ write messages onto the queue:
 
 ```ruby
 message_data = {
-  "event_type" => "heartbeat",
   "data" => {
```
alieander, line 55: Data seems almost unneeded anymore... but we should keep it as it helps us conform to the form of JsonApi we are looking for in other places. I think I could be beneficial to continue to conform to that even in the formatting of our messages. As a good start, `"data"` is a top level item in our JsonApi world.

[JsonApi.org](http://jsonapi.org/)
[Fastly fst-json_api](https://github.com/fastly/fst-json_api)

It might be good to think about what we want the body of our messages to conform to before we have to many.

examples/Rakefile
```diff
@@ -1,41 +1,111 @@
 require 'fastly_nsq'
 require 'fastly_nsq/rake_task'
```
alieander: You commented that one of these is not needed? We should remove it from here as well.

examples/Rakefile
```diff
@@ -1,41 +1,111 @@
...
-# assigning value like so. This is useful if you want a single purpose
-# listener task.
+#   # ...
+# end
 #
-# ex: `rake listen_task`
-# topic will be 'some_topic' and channel will be 'some_channel'
+# A channel will need to be passed in
+# to the Rake task
+# when it is called.
```
alieander: Why so many new lines here?

examples/Rakefile
```diff
@@ -1,41 +1,111 @@
...
 #
-MessageQueue::RakeTask.new(:listen_task) do |task|
-  task.topic   = 'some_topic'
-  task.channel = 'some_channel'
+# The task looks like:
+
+namespace :queue do
```
alieander: Do we really need this namespace here? It does not need to be around for this to function and feels like it is unnecessary.

examples/Rakefile
```diff
@@ -1,41 +1,111 @@
...
-# In Rails, you can include the environment like so:
-#
-MessageQueue::RakeTask.new(:listen_task, [:topic, :channel] => :environment)
-
-##
-# Topic and channel can also be preset in the task by passing a block and
-# assigning value like so. This is useful if you want a single purpose
-# listener task.
+#   # ...
+# end
```
alieander: Have you considered uncommenting this as it would still be valid ruby and makes some things later make more sense when you reference what the topics will become.

examples/Rakefile
```diff
@@ -1,41 +1,111 @@
...
 #
-MessageQueue::RakeTask.new(:listen_task) do |task|
-  task.topic   = 'some_topic'
-  task.channel = 'some_channel'
+# The task looks like:
+
+namespace :queue do
+  desc 'Listen to the messaging queue with the given channel.'
+
+  MessageQueue::RakeTask.new([:channel])
```
alieander: Have you tried this in this form? I thought the first argument had to be the `name` of the task... I could be wrong.

lib/fastly_nsq/rake_task.rb
```diff
@@ -3,36 +3,71 @@
...
+      output "Listening to queue, topic:'#{topic}' and channel: '#{channel}'"
+      yield
+      output "... done listening on topic:'#{topic}' and channel: '#{channel}'."
+    end
+
+    def topics
+      MessageProcessor.topics
+    rescue NoMethodError => exception
+      if exception.message =~ /undefined method `topics'/
+        raise ArgumentError, "MessageProcessor.topics is not defined."
```
alieander: Rubocop might not like the double quotes here.

lib/fastly_nsq/rake_task.rb
```diff
@@ -3,36 +3,71 @@
...
+      unless channel
+        raise ArgumentError, "channel is required. Received channel: #{channel}"
+      end
+    end
+
+    def wrap_helpful_output(topic)
+      output "Listening to queue, topic:'#{topic}' and channel: '#{channel}'"
+      yield
+      output "... done listening on topic:'#{topic}' and channel: '#{channel}'."
+    end
```
alieander, line 59: Since we have the concept of a logger now, might we also want `output` to look at using that?

lib/fastly_nsq/message_queue/listener.rb
```diff
@@ -26,14 +26,14 @@ def process_next_message
 
     private
 
+    attr_reader :channel, :topic
+
     def process_one_message
       message = consumer.pop
-      MessageProcessor.new(message.body).go
+      MessageProcessor.new(topic: topic,  message_body: message.body).go
```
alieander: I know with keyword args the order is not important... but it mildly bothers me that this is backwards from how it is defined below in the sample message processor

lib/fastly_nsq/version.rb
```diff
@@ -1,3 +1,3 @@
 module FastlyNsq
-  VERSION = '0.2.3'
+  VERSION = '0.3.0'
```
alieander, line 2: :yellow_heart:

spec/lib/fastly_nsq/rake_task_spec.rb
```diff
@@ -3,71 +3,124 @@
...
+      end
     end
   end
 
   describe 'when running tasks' do
-    it 'runs with inline options defined' do
-      options = { topic: 'dwarf', channel: 'star' }
-      task = MessageQueue::RakeTask.new(:begin_listening, [:topic, :channel])
+    context 'when multiple topics are defined' do
+      it 'listens to both channels' do
```
alieander: I think you mixed up `channel` and `topic` a bit in the wording here.
Additionally, it doesn't look to me like this goes out of its whay to check that the channel is correct or even being checked...
Perhaps:

``` ruby
context 'when multiple topics are defined' do
  it 'listens to both on the same chanel' do
```

or:

``` ruby
context 'when multiple topics are defined' do
  it 'creates a listener for each' do
```

alieander:
Left a bunch of comments!
Looks great overall! Well done Sir!

lib/fastly_nsq/rake_task.rb
```diff
@@ -3,36 +3,71 @@
...
-          run_task
+          if block_given?
+            yield(*[self, task_args].slice(0, task_block.arity))
+          end
+
+          if task_args[:channel]
+            @channel = task_args[:channel]
+          end
+
+          guard_missing_channel
```
standingwave, line 25: I like to guard early for missing required args. Potentially the block above this could also throw an exception because of a bad or missing channel.

README.md
adarsh, line 55: > I think I could be beneficial to continue to conform to that even in the formatting of our messages. As a good start, "data" is a top level item in our JsonApi world.

Oh nice, I totally agree. I'm going to keep the top-level `data` key for now. Might be worth discussing a follow-up PR to ensure it conforms. 

Funny, I actually didn't think to add `data:` as a key because of JSON-API, but looks like it was a happy coincidence :)

examples/Rakefile
adarsh: Yeah the general `require 'fastly_nsq'` I believe is not required - we only want the rake task here, and it's only dependency is on `MessageProcessor`. I've removed in 1d23dd7

examples/Rakefile
adarsh: ### Short answer

"Semantic linefeeds", which means "one line per thought/clause". The goal is to reduce diffs, much like the trailing comma in a multi-line Ruby list. Also it separates ideas (for me) and is easier to write prose properly (again, for me). Markdown doesn't care and will jam them all into one paragraph.

### Long answer

The above, plus this is something which goes back to the Bell Labs UNIX philosophy of dealing only in plain text:

> Hints for Preparing Documents
> 
> Most documents go through several versions (always more than you expected) before they are finally finished. Accordingly, you should do whatever possible to make the job of changing them easy.
> 
> First, when you do the purely mechanical operations of typing, type so subsequent editing will be easy. Start each sentence on a new line. Make lines short, and break lines at natural places, such as after commas and semicolons, rather than randomly. Since most people change documents by rewriting phrases and adding, deleting and rearranging sentences, these precautions simplify any editing you have to do later.
> 
> — Brian W. Kernighan, Bell Labs, 1974
> From "UNIX for Beginners" (http://miffy.tom-yam.or.jp/2238/ref/beg.pdf)

examples/Rakefile
adarsh: We certainly don't need it in this example file. Removed in f53aebc.

It is hardwired into the listening cookbooks, so those will need updating if we change in the Rails apps, but they need it anyhow so it would be a good time to adjust. :+1:

examples/Rakefile
adarsh: Oh good point, nice! Also, this takes advantage of syntax highlighting for the reader.

Done in 7aac3e4.

examples/Rakefile
adarsh: I have not, how negligent of me! It does not work, so I removed this example in 2450902. I'll check the rest now as well.

adarsh: All the rest of the examples work as described. Good catch!

lib/fastly_nsq/rake_task.rb
adarsh: Whoops, looks like no `.rubocop.yml` file or `.overcommit.yml`. I will fix in a subsequent PR.

Fixed the quotes here cce2a44.

lib/fastly_nsq/rake_task.rb
adarsh, line 59: Yes! The output is much better:

```
Listening to queue, topic:'heartbeat' and channel: 'testing_stuff'
Listening to queue, topic:'fastly_customer_created' and channel: 'testing_stuff'
```

vs

```
I, [2016-03-17T13:04:43.286756 #48125]  INFO -- : Listening to queue, topic:'heartbeat' and channel: 'william'
I, [2016-03-17T13:04:43.286872 #48125]  INFO -- : Listening to queue, topic:'fastly_customer_created' and channel: 'william'
```

Done in 40f079c.

lib/fastly_nsq/message_queue/listener.rb
adarsh: OMG how embarrassing. Fixed here and elsewhere in 8e65433, only for aesthetic reasons.

lib/fastly_nsq/version.rb
adarsh, line 2: This feels like a big enough change that it warrants a minor version bump, I assume you concur - I'm happy to discuss if otherwise.

spec/lib/fastly_nsq/rake_task_spec.rb
adarsh: Oops, good catch, thanks! Fixed in ae32027, I liked the second wording.

lib/fastly_nsq/rake_task.rb
adarsh, line 25: Agree: I like to as well. Sort of the point of a guard, right?

At any rate, moving this line up even one block breaks a tests for some reason. This code is a little hard to follow, I'd welcome suggestions for how to fix the test or the code.

alieander:
I have re-reviewed. Seems Great!

adarsh:
Thanks for thorough review @alieander!

On Thu, Mar 17, 2016 at 2:05 PM, alieander notifications@github.com wrote:

> I have re-reviewed. Seems Great!
> 
> —
> You are receiving this because you authored the thread.
> https://github.com/fastly/fastly_nsq/pull/22#issuecomment-198082246

## 

Adarsh Pandit
External Consultant
+1 313 454 1515 (m)

PR #23, opened 2016-03-18
Add Rubocop and Overcommit

adarsh:
# Reason for Change
- We would like to enforce style guidelines in our application programmatically.
# Changes
- Add the Rubocop gem and `rake rubocop` task.
- Import the Rubocop config file from another app.
- Fix Rubocop violations.
- Enforce Rubocop using Overcommit post-git hooks via it's gem and `.overcommit.yml` config file.
- Have Rubocop target Ruby 2.1, since that is the lowest supported version.
- Reduce branch complexity of a method
## Minor
- Resort Rake tasks to match the order in which they run.
## Risks
- None, these are development enhancements.

Done as part of this Jira but PRed separately:
https://jira.corp.fastly.com/browse/HYD-735
# Checklist

# Recommended Reviewers

@alieander, @standingwave

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -1,0 +1,34 @@
+PreCommit:
+  ALL:
+    quiet: true
+    verify_signatures: false
+
+  MergeConflicts:
+    description: 'Checking for unaddressed merge conflicts'
+    enabled: true
+    quiet: true
+    required_executable: 'grep'
+    flags: ['-IHn', "^<<<<<<<[ \t]"]
+
+  RuboCop:
+    description: 'Checking Ruby style'
+    enabled: true
+    problem_on_unmodified_line: ignore
+    quiet: true
+    on_warn: fail
+
+CommitMsg:
+  ALL:
+    quiet: true
+  CapitalizedSubject:
+    description: 'Ensuring commit message subjects are capitalized'
+    enabled: false
+  EmptyMessage:
+    description: 'Prevent empty commit messages'
+    enabled: true
+  TextWidth:
+    description: 'Ensure commit message subjects are <60 chars'
+    enabled: true
+    description: 'Checking text width'
+    max_subject_width: 60
+    max_body_width: 1000
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,72 @@
+AllCops:
+  Include:
+    - '**/Rakefile'
+  Exclude:
+    - 'db/schema.rb'
+    - 'vendor/**/*'
+  TargetRubyVersion: 2.1
+
+LineLength:
+  Enabled: false
+
+HashSyntax:
+  Enabled: false
+
+BracesAroundHashParameters:
+  Enabled: false
+
+CollectionMethods:
+  Enabled: false
+
+CyclomaticComplexity:
+  Enabled: true
+
+Style/DotPosition:
+  Enabled: true
+  EnforcedStyle: trailing
+  SupportedStyles:
+    - leading
+    - trailing
+
+MethodLength:
+  Enabled: false
+
+PerlBackrefs:
+  Enabled: false
+
+RescueModifier:
+  Enabled: false
+
+GuardClause:
+  Enabled: false
+
+IfUnlessModifier:
+  Enabled: false
+
+ClassLength:
+  Enabled: false
+
+Style/ClassVars:
+  Description: 'Avoid the use of class variables.'
+  StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-class-vars'
+  Enabled: false
+
+Documentation:
+  Enabled: false
+
+RegexpLiteral:
+  Enabled: false
+
+SignalException:
+  EnforcedStyle: only_raise
+
+Style/ClassAndModuleChildren:
+  Enabled: false
+
+Style/TrailingCommaInLiteral:
+  EnforcedStyleForMultiline: comma
+  Enabled: true
+
+Style/TrailingCommaInArguments:
+  EnforcedStyleForMultiline: comma
+  Enabled: true
--- a/spec/lib/fastly_nsq/rake_task_spec.rb
+++ b/spec/lib/fastly_nsq/rake_task_spec.rb
@@ -37,7 +37,7 @@ RSpec.describe MessageQueue::RakeTask do
     context 'when multiple topics are defined' do
       it 'creates a listener for each' do
         channel = 'clown_generating_service'
-        topics = ['customer_created', 'customer_now_awesome']
+        topics = %w(customer_created customer_now_awesome)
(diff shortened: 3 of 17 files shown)


.overcommit.yml
```diff
@@ -0,0 +1,34 @@
...
+    enabled: false
+  EmptyMessage:
+    description: 'Prevent empty commit messages'
+    enabled: true
+  TextWidth:
+    description: 'Ensure commit message subjects are <60 chars'
+    enabled: true
+    description: 'Checking text width'
+    max_subject_width: 60
+    max_body_width: 1000
```
alieander, line 34: What if I want to write a small novel?

alieander, line 34: This is a joke...

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+  Enabled: false
+
+AssignmentInCondition:
+  Enabled: false
+
+CollectionMethods:
+  Enabled: false
+
+CyclomaticComplexity:
+  Enabled: false
```
alieander: This should not be disabled...

alieander: Better comment: why is this disabled?

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
+AllCops:
+  Include:
+    - '**/Rakefile'
+  Exclude:
+    - 'bin/stubs/*'
+    - 'cookbooks/**/*'
```
alieander: This file should now be not have any ruby in it anymore. It can be removed from here

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
+AllCops:
+  Include:
+    - '**/Rakefile'
+  Exclude:
+    - 'bin/stubs/*'
```
alieander: I thought this was specific to personal setup?

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+    - 'cookbooks/**/*'
+    - 'db/schema.rb'
+    - 'vendor/**/*'
+  TargetRubyVersion: 2.1
```
alieander: What does this do? And why 2.1 and not 2.3?

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+  Enabled: false
+
+HashSyntax:
+  Enabled: false
+
+BracesAroundHashParameters:
+  Enabled: false
+
+BlockNesting:
+  Enabled: false
```
alieander: This should not be disabled

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+
+AssignmentInCondition:
+  Enabled: false
```
alieander: This one could be talked about as well...

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+
+PerlBackrefs:
+  Enabled: false
+
+RescueModifier:
+  Enabled: false
+
+HandleExceptions:
+  Enabled: false
```
alieander: This, when on, keeps people from rescuing and then doing nothing.

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+
+Style/TrailingCommaInLiteral:
+  Enabled: false
+
+Style/TrailingCommaInArguments:
+  Enabled: false
+
+UnusedMethodArgument:
+  Enabled: false
```
alieander: This one I kinda like having on...

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+  EnforcedStyle: only_raise
+
+Style/ClassAndModuleChildren:
+  Enabled: false
+
+Style/TrailingCommaInLiteral:
+  Enabled: false
+
+Style/TrailingCommaInArguments:
+  Enabled: false
```
alieander: This should not be off. One of the options should be set. I think having it off allows `def thing(arg, arg,)`
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/TrailingCommaInArguments

.rubocop.yml
```diff
@@ -0,0 +1,84 @@
...
+
+SignalException:
+  EnforcedStyle: only_raise
+
+Style/ClassAndModuleChildren:
+  Enabled: false
+
+Style/TrailingCommaInLiteral:
+  Enabled: false
```
alieander: Same as below.

spec/lib/fastly_nsq/rake_task_spec.rb
```diff
@@ -37,7 +37,7 @@
     context 'when multiple topics are defined' do
       it 'creates a listener for each' do
         channel = 'clown_generating_service'
-        topics = ['customer_created', 'customer_now_awesome']
+        topics = %w(customer_created, customer_now_awesome)
```
alieander: The comma here is not needed and might be a part of the first string...

alieander:
Finished. Lots of comments on the RoboCop preferences. We should talk about these face to face.

.rubocop.yml
adarsh: Unclear, I've enabled it in 4d20b6f

.rubocop.yml
adarsh: Agree. Done in c25c02e

.rubocop.yml
adarsh: It is, removed. 8e8c293

.rubocop.yml
adarsh: Short answer: we support Ruby >= 2.1 so that's why this is our target.

As for what it does:

> Some checks are dependent on the version of the Ruby interpreter which the inspected code must run on. For example, using Ruby 2.0+ keyword arguments rather than an options hash can help make your code shorter and more expressive... unless it must run on Ruby 1.9.

https://github.com/bbatsov/rubocop#setting-the-target-ruby-version

.rubocop.yml
adarsh: Agreed. It is enabled by default, so I deleted it in e565c2d

.rubocop.yml
adarsh: As discussed IRL, it should be on, and is by default so removed this in 2795027.

.rubocop.yml
adarsh: Agree. 629c711

.rubocop.yml
adarsh: Agree 5cd044c

.rubocop.yml
adarsh: I agree! c4e4eae

spec/lib/fastly_nsq/rake_task_spec.rb
adarsh: Fixed in 40e9638.

alieander:
This is GREAT!!!

PR #25, opened 2016-04-15
Add SSLContext to fastly_nsq

alieander:
# Reason for change

We would like to be able to use TLS/SSL.
This allows us to pass a SSLContext thru the producer or consumer to the
nsq-ruby gem which will then use that information to encrypt communitcations.
We would also like for this to be easily setup via ENV variables.
# Changes
- Add an SSLContext class with tests
- Use this class in Consumer and Producer
- Allow consumer and producer to accept a passed in hash ssl_context
- Forward this SSLContext to the underlying Strategy
- Bump nsq-ruby gem version to 1.6.0

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
+gem 'nsq-ruby', '~> 1.5.0', '>= 1.5.0', git: 'https://github.com/fastly/nsq-ruby.git'
+
 gemspec
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,3 @@
 source 'https://rubygems.org'
 
-gem 'nsq-ruby', '~> 1.5.0', '>= 1.5.0', git: 'https://github.com/fastly/nsq-ruby.git'
-
 gemspec
--- a/lib/fastly_nsq/message_queue/consumer.rb
+++ b/lib/fastly_nsq/message_queue/consumer.rb
@@ -2,9 +2,10 @@ class InvalidParameterError < StandardError; end
 
 module MessageQueue
   class Consumer
-    def initialize(topic:, channel:)
+    def initialize(topic:, channel:, ssl_context: nil)
       @topic = topic
       @channel = channel
+      @ssl_context = SSLContext.new(ssl_context)
     end
 
     def connection
@@ -17,7 +18,7 @@ module MessageQueue
 
     private
 
-    attr_reader :channel, :topic
+    attr_reader :channel, :topic, :ssl_context
 
     def consumer
       Strategy.for_queue::Consumer
@@ -28,6 +29,7 @@ module MessageQueue
         nsqlookupd: ENV.fetch('NSQLOOKUPD_HTTP_ADDRESS'),
         topic: topic,
         channel: channel,
+        ssl_context: ssl_context.to_h,
       }
     end
   end
--- a/lib/fastly_nsq/ssl_context.rb
+++ b/lib/fastly_nsq/ssl_context.rb
@@ -1,0 +1,44 @@
+class SSLContext
+  def initialize(context = nil)
+    @context = context || {}
+  end
+
+  def to_h
+    merge_contexts
+    if empty_context?
+      nil
+    else
+      @context
+    end
+  end
+
+  private
+
+  def env_key
+    ENV.fetch('NSQ_SSL_KEY', nil)
+  end
+
+  def env_certificate
+    ENV.fetch('NSQ_SSL_CERTIFICATE', nil)
+  end
+
+  def env_ca_certificate
+    ENV.fetch('NSQ_SSL_CA_CERTIFICATE', nil)
+  end
+
+  def env_default_hash
+    {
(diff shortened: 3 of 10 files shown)


Gemfile
```diff
@@ -1,3 +1,5 @@
 source 'https://rubygems.org'
 
+gem 'nsq-ruby', '~> 1.5.0', '>= 1.5.0', git: 'https://github.com/fastly/nsq-ruby.git'
+
```
adarsh: This seems like a fine change:

> If you find that you need to develop against a gem that hasn't yet been released (for instance, Rails often develops against unreleased Rack, Arel, or Mail gems), you can add individual lines in the Gemfile that tell bundler where to find the gems. It will still use the dependencies listed in the .gemspec for dependency resolution, but it now knows exactly where to find the dependency during gem development.

http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

lib/fastly_nsq/message_queue/consumer.rb
```diff
@@ -2,9 +2,10 @@ class InvalidParameterError < StandardError; end
 
 module MessageQueue
   class Consumer
-    def initialize(topic:, channel:)
+    def initialize(topic:, channel:, ssl_context: nil)
       @topic = topic
       @channel = channel
+      @ssl_context = SSLContext.new(ssl_context)
```
adarsh, line 8: I really like this.

lib/fastly_nsq/ssl_context.rb
```diff
@@ -0,0 +1,44 @@
...
+  def env_default_hash
+    {
+      key: env_key,
+      certificate: env_certificate,
+      ca_certificate: env_ca_certificate,
+    }
+  end
+
+  def merge_contexts
+    @context = env_default_hash.merge!(@context)
```
adarsh: Consider using a new ivar here instead of mutating the existing one, outside chance this could lead to collisions or confusion.

lib/fastly_nsq/ssl_context.rb
```diff
@@ -0,0 +1,44 @@
+class SSLContext
+  def initialize(context = nil)
+    @context = context || {}
```
adarsh, line 3: I might be wrong but I believe you can use a NullObject here, like:

``` ruby
@context = context || NullSSLContext.new
```

And have that just respond to `.to_h` with a `nil`.

spec/lib/fastly_nsq/ssl_context_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'spec_helper'
+
+RSpec.describe SSLContext do
+  describe 'when SSL ENV variables are not set' do
+    describe '.to_h' do
+      it 'returns nil when initilized without parameters' do
```
adarsh: Typo: 'initilized'

spec/lib/fastly_nsq/ssl_context_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        context = SSLContext.new
+
+        expect(context.to_h).to be_nil
+      end
+
+      it 'returns a hash the same hash when all variables are defined' do
```
adarsh: This reads a little funky, maybe a typo? "returns a hash the same hash".

spec/lib/fastly_nsq/ssl_context_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        ca_certificate = context.to_h[:ca_certificate]
+
+        expect(ca_certificate).to be_nil
+      end
+    end
+  end
+
+  describe 'when SSL ENV variables are set' do
+    before do
+      ENV['NSQ_SSL_KEY'] = '/some/key'
```
adarsh, line 38: So these changes will persist across tests when the `ENV` is mutated.

Instead, consider mocking the behavior out, as we do [here](https://github.com/fastly/fastly_nsq/blob/master/spec/support/env_helpers.rb#L4):

``` ruby
allow(ENV).to receive(:[]).with('FAKE_QUEUE').and_return(yes)
```

adarsh:
Minor comments, looks 👍  to me.

spec/lib/fastly_nsq/ssl_context_spec.rb
alieander, line 38: sure

lib/fastly_nsq/ssl_context.rb
leklund: The `!` is unnecessary

spec/lib/fastly_nsq/ssl_context_spec.rb
alieander, line 38: allow ENV doesn't seem to work well with multiple things needing to be mocked out.

lib/fastly_nsq/ssl_context.rb
alieander, line 3: This would drastically change how the rest of this class works.

PR #26, opened 2016-04-19
Should Listener#consumer -> Consumer, or Consumer#connection?

jaw6:
`Listener#consumer` attaches to `MQ::Consumer#connection` instead of `MQ::Consumer`. I'm :new: here, but I _think_ this means that we have the actual `nsq-ruby` object inside `Listener`, instead of the `MQ::Consumer` wrapper. And, possibly tangentially, that `Listener` doesn't actually respect the `fake` queue `Strategy` (since `FakeMessageQueue::Consumer` doesn't implement `connection`)

This change breaks some tests. I thought I'd check in and see if I've wildly misunderstood any of this, before I stumbled around too much in fixing these.

---

So: I continued to run with the idea that Consumer should wrap the under NSQ Consumer, and that, in effect, MessageQueue::Consumer#connection ought to be private. Made tests pass. Along the way, some refactoring occurred, mostly in tests.

Assuming review shakes out, this should probably also bump the gem version, since we may need time to get downstream services weaned off of using #connection directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,14 @@
 # fastly_nsq [![Build Status](https://travis-ci.org/fastly/fastly_nsq.svg?branch=master)](https://travis-ci.org/fastly/fastly_nsq)
 
-*NOTE: This is a point-release
-which is not yet suitable for production.
-Use at your own peril.*
+*NOTE: This is a point-release which is not yet suitable for production. Use at your own peril.*
 
-NSQ adapter and testing objects
-for using the NSQ messaging system
-in your Ruby project.
+NSQ adapter and testing objects for using the NSQ messaging system in your Ruby project.
 
-This library is intended
-to facilitate publishing and consuming
-messages on an NSQ messaging queue.
+This library is intended to facilitate publishing and consuming messages on an NSQ messaging queue.
 
-We also include fakes
-to make testing easier.
+We also include fakes to make testing easier.
 
-This library is dependent
-on the [`nsq-ruby`] gem.
+This library is dependent on the [`nsq-ruby`] gem.
 
 [`nsq-ruby`]: https://github.com/wistia/nsq-ruby
 
@@ -27,28 +19,18 @@ Please use [GitHub Issues] to report bugs.
 
 ## Install
 
-`fastly_nsq` is a Ruby Gem
-tested against Rails `>= 4.2`
-and Ruby `>= 2.3.0`.
+`fastly_nsq` is a Ruby Gem tested against Rails `>= 4.2` and Ruby `>= 2.3.0`.
 
-To get started,
-add `fastly_nsq` to your `Gemfile`
-and `bundle install`.
+To get started, add `fastly_nsq` to your `Gemfile` and `bundle install`.
 
 ## Usage
-
-*IMPORTANT NOTE:* You must create your own `MessageProcessor` class
-for this gem to work in your application.
+*IMPORTANT NOTE:* You must create your own `MessageProcessor` class for this gem to work in your application.
 
 See more information below.
 
 ### `MessageQueue::Producer`
 
-This is a class
-which provides an adapter to the
-fake and real NSQ producers.
-These are used to
-write messages onto the queue:
+This is a class which provides an adapter to the fake and real NSQ producers. These are used to write messages onto the queue:
 
 ```ruby
 message_data = {
@@ -64,10 +46,8 @@ producer = MessageQueue::Producer.new(
 
 producer.write(message_data.to_json)
 ```
-The mock/real strategy used
-can be switched
-by adding an environment variable
-to your application:
+
+The mock/real strategy used can be switched by adding an environment variable to your application:
 
 ```ruby
 # for the fake
@@ -78,11 +58,8 @@ ENV['FAKE_QUEUE'] = false
 ```
 
 ### `MessageQueue::Consumer`
-This is a class
-which provides an adapter to the
-fake and real NSQ consumers.
-These are used to
-read messages off of the queue:
+
+This is a class which provides an adapter to the fake and real NSQ consumers. These are used to read messages off of the queue:
 
 ```ruby
 consumer = MessageQueue::Consumer.new(
@@ -98,10 +75,7 @@ consumer.size #=> 0
 consumer.terminate
 ```
 
-As above,
-the mock/real strategy used
-can be switched by setting the
-`FAKE_QUEUE` environment variable appropriately.
+As above, the mock/real strategy used can be switched by setting the `FAKE_QUEUE` environment variable appropriately.
 
 ### `MessageQueue::Listener`
 
@@ -114,10 +88,7 @@ channel = 'my_consuming_service'
 MessageQueue::Listener.new(topic: topic, channel: channel).process_next_message
 ```
 
-This will pop the next message
-off of the queue
-and send the JSON text body
-to `MessageProcessor.new(message_body: message_body, topic: topic).go`.
+This will pop the next message off of the queue and send the JSON text body to `MessageProcessor.new(message_body: message_body, topic: topic).go`.
 
 To initiate a blocking loop to process messages continuously:
 
@@ -128,26 +99,15 @@ channel = 'my_consuming_service'
 MessageQueue::Listener.new(topic: topic, channel: channel).go
 ```
 
-This will block until
-there is a new message on the queue,
-      pop the next message
-      off of the queue
-      and send it to `MessageProcessor.new(message_body).go`.
+This will block until there is a new message on the queue, pop the next message off of the queue and send it to `MessageProcessor.new(message_body).go`.
 
 ### `MessageQueue::RakeTask`
 
-To help facilitate running the `MessageQueue::Listener` in a blocking fashion
-outside your application, a simple `RakeTask` is provided.
+To help facilitate running the `MessageQueue::Listener` in a blocking fashion outside your application, a simple `RakeTask` is provided.
 
-NOTE: The rake task expects a
-`MessageProcessor.topics` method,
-which must return an array of strings
-defining the topics to which
-we would like to listen and process messages.
+NOTE: The rake task expects a `MessageProcessor.topics` method, which must return an array of strings defining the topics to which we would like to listen and process messages.
 
-The task will listen
-to all specified topics,
-each in a separate thread.
+The task will listen to all specified topics, each in a separate thread.
 
 This task can be added into your `Rakefile` in one of two ways:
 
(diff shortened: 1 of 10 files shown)


adarsh:
Nice work! PR on the 2nd day and great questions!

So:

> Listener#consumer attaches to MQ::Consumer#connection instead of MQ::Consumer. I'm :new: here, but I think this means that we have the actual nsq-ruby object inside Listener, instead of the MQ::Consumer wrapper.

Yes, this is correct and the desired behavior - an open connection which is terminated in a few different situations (`SIGTERM`, only processing one message, etc.)

I'm guessing the confusion comes from the method name? Would a renaming of this private method from `listener` to `listener_connection` help with clarity? If so, I would be in favor of that.

> And, possibly tangentially, that Listener doesn't actually respect the fake queue Strategy (since FakeMessageQueue::Consumer doesn't implement connection)

This is more complicated, since the Fake Strategy actually never opens a connection to anything other than an array which serves as the stack for FIFO work.

You may be right, though. Can you provide some more color around a possible strategy mismatch and/or how we might resolve it?

jaw6:
> I'm guessing the confusion comes from the method name?

Partly. I think a clearer method name would better reflect the _current_ behavior, but...

> Can you provide some more color around a possible strategy mismatch and/or how we might resolve it?

This is probably me coming in with some assumptions, but I was expecting `fastly_nsq` to provide more of an abstract wrapper. Because we're leaking the underlying `connection`, we can't leverage `fastly_nsq` classes to add additional/common behavior. (Or, in this case, leverage a non-live backend.)

Or: if I were working to implement `fastly_nsq` in some other service, I think I'd expect I'd be able to drive dev/test environments using `FakeMessageQueue` (at least, up to some point). I'd set the appropriate `ENV` vars and my `Listener` tests wouldn't work as I expected.

adarsh:
As discussed IRL, a delegator may help this problem.

On Wed, Apr 20, 2016 at 11:28 AM, Joshua Wehner notifications@github.com
wrote:

> I'm guessing the confusion comes from the method name?
> 
> Partly. I think a clearer method name would better reflect the _current_
> behavior, but...
> 
> Can you provide some more color around a possible strategy mismatch and/or
> how we might resolve it?
> 
> This is probably me coming in with some assumptions, but I was expecting
> fastly_nsq to provide more of an abstract wrapper. Because we're leaking
> the underlying connection, we can't leverage fastly_nsq classes to add
> additional/common behavior. (Or, in this case, leverage a non-live backend.)
> 
> Or: if I were working to implement fastly_nsq in some other service, I
> think I'd expect I'd be able to drive dev/test environments using
> FakeMessageQueue (at least, up to some point). I'd set the appropriate ENV
> vars and my Listener tests wouldn't work as I expected.
> 
> —
> You are receiving this because you were mentioned.
> https://github.com/fastly/fastly_nsq/pull/26#issuecomment-212546438

## 

Adarsh Pandit
External Consultant
+1 313 454 1515 (m)

jaw6:
@fastly/billing the work here trickled in over the course of my first week, but I don't think I knew enough to, say, attach labels. At any rate, does anyone have time to give this a more official review/sign-off?

alieander:
Let me take a quick look.

alieander:
I like all the changes.
It seems like an added gain to allow a MessageProcessor to be passed in if the implementer wants to do that. Are there changes to the README that could reflect that ability?

spec/lib/fastly_nsq/fake_message_queue_spec.rb
```diff
@@ -88,65 +88,55 @@
 end
 
 RSpec.describe FakeMessageQueue::Consumer do
+  let(:topic)   { 'death_star' }
+  let(:channel) { 'star_killer_base' }
+  let(:consumer) { FakeMessageQueue::Consumer.new topic: topic, channel: channel }
```
alieander, line 93: alignment?

jaw6, line 93: Good catch!

leklund:
for travis and broken bundler. Update the gemspec to use `bundler ~> 1.12.2` ? maybe?

jaw6:
So... this is annoying, but as far as I can tell, there aren't better options.

As I understand it:
1. In _theory_, `before_install: gem install bundler -v '1.11.2'` is _supposed_ to work, but, at the moment, doesn't actually.
   - To be clear: it _does_ install `bundler -v 1.11.2` but the next `bundle install` activates the wrong bundler (`1.12.2`) regardless
2. This isn't just us, but it seems to be mostly folks trying to run _newer_ bundler and getting an unexpectedly old Bundler: [Travis CI](https://github.com/travis-ci/travis-ci/issues/5798) but also [other Ruby-CI-type folks](https://discuss.circleci.com/t/bundler-fails-to-find-appropriate-version-despite-installing-appropriate-version-earlier-in-the-build/2815/35) have been discovering that they couldn't run _newer_ Bundler
3. About a month ago, Rubygems had a problem where it would choose the oldest available gem
   - This [has been fixed](https://github.com/rubygems/rubygems/pull/1566) and is now available in rubygems / `gem >= 2.6.4`
4. Oddly, `rvm` includes it's own version of `bundler` which exists outside of the ruby gemset (see image below)
5. Somewhere in the last ~week (?) Travis updated one or more of `[ rvm rubygems bundler ]` and
6. It is currently not possible to lock a `Gemfile / gemspec` to a version of Bundler that is not `1.12.2` (or, presumably, `bundler $current` at some point in the near future)

screen shot 2016-05-05 at 4 49 46 pm

Possible solutions available to us:
1. Continue to search for a way to tell `rvm` / `travis` / `bundler` to run `bundle` in some way that targets the "correct" bundler (eg, `/path/to/rvm/gems/1.11.2/bin/bundle install` and `/path/to/rvm/gems/1.11.2/bin/bundle exec rake` and so on)
2. Remove or broaden the explicit bundler version lock in our `gemspec` - do we _really_ need this version of bundler for some reason?
3. Update the bundler version we lock to whenever `travis` / `rvm` / `bundler` lurches forward

**_Right now**_, I'm using option 3 (update our locked bundler) but I'd probably rather do option 2 if we can (eg, either `bundler ~> 1` or ~~bundler~~)

Thoughts from @fastly/billing ? Anybody know/remember/guess why we locked to this version of bundler in the first place?

fastly_nsq.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ['lib']
 
   gem.add_development_dependency 'awesome_print', '~> 1.6'
-  gem.add_development_dependency 'bundler', '~> 1.11.2'
+  gem.add_development_dependency 'bundler', '~> 1.12'
```
drbrain, line 24: The best way is `'~> 1.12'` as written since it allows upgrades to compatible future versions.  If we need to be at a particular bug fix level, `'~> 1.12', '>= 1.12.2'` would allow us to use newer compatible versions.

lib/fastly_nsq/message_queue/consumer.rb
```diff
@@ -1,27 +1,26 @@
+require 'forwardable'
+
 class InvalidParameterError < StandardError; end
 
 module MessageQueue
   class Consumer
+    extend Forwardable
+    def_delegator :connection, :pop
+    def_delegator :connection, :pop_without_blocking
+    def_delegator :connection, :terminate
```
leklund, line 11: This is beautiful.

leklund:
I think this looks great! Can you update the README with the new usage?

README.md
```diff
@@ -242,36 +184,21 @@ NSQLOOKUPD_TCP_ADDRESS='127.0.0.1:4160'
...
-`FakeMessageQueue.reset!`
-before each test in your app to ensure
-there are no leftover messages.
-
-Also note that during gem tests,
-we are aliasing `MessageProcessor` to `SampleMessageProcessor`.
-You can also refer to the latter
-as an example of how
-you might write your own processor.
+Also note that during gem tests, we are aliasing `MessageProcessor` to `SampleMessageProcessor`. You can also refer to the latter as an example of how you might write your own processor.
```
adarsh: I realize this format of linebreaking in the `README.md` seems in error, but was actually intentional [semantic linefeeds](http://rhodesmill.org/brandon/2012/one-sentence-per-line/).

I'm good with the change, if you and others find it easier to read/work with.

lib/fastly_nsq/fake_message_queue.rb
```diff
@@ -42,22 +52,24 @@ def queue
...
     end
 
-    def pop
+    def pop(delay = FakeMessageQueue.delay)
       message = nil
 
       until message
         message = queue.pop
-        sleep SECONDS_BETWEEN_QUEUE_CHECKS
+        sleep delay
```
adarsh, line 63: Nice!

lib/fastly_nsq/message_queue/listener.rb
```diff
@@ -1,8 +1,10 @@
 module MessageQueue
   class Listener
-    def initialize(topic:, channel:)
-      @topic = topic
-      @channel = channel
+    def initialize(topic:, channel:, processor: nil, consumer: nil)
+      @topic     = topic
+      @channel   = channel
+      @processor = processor || DEFAULT_PROCESSOR
```
adarsh, line 6: I might be wrong but I believe you can do

``` ruby
def initialize(topic:, channel:, processor: DEFAULT_PROCESSOR, consumer: nil)
  @processor = processor
  #...
```

and have the same effect.

adarsh:
I'm a fan 👍

README.md
jaw6: My gut says this is non-standardized over-optimizing for `diff` and I'd prefer to optimize for the reader - one of markdown's selling points is that it's fairly readable as plain text, even without rendering. But if this style is popular here, I'm sure I can adapt. 

I'm curious where @alieander, @leklund and @standingwave land on this.

lib/fastly_nsq/message_queue/listener.rb
jaw6, line 6: That's true. Is it more readable/intention-revealing in that form? I think, long-term, I'd like to see us move away from the "magic constant" `MessageProcessor` in favor of this kind of configuration (maybe with some sane defaults), so if we need to establish how we're going to write that sort of thing now, maybe it's a good time.

adarsh, line 6: Right. Actually, on re-reading, your implementation is clearer,
particularly because it's broken out into a separate line. This is
particularly true with the next assignment down, which is more involved.

Please ignore my comment then.

On Mon, May 9, 2016 at 9:56 AM, Joshua Wehner notifications@github.com
wrote:

> In lib/fastly_nsq/message_queue/listener.rb
> https://github.com/fastly/fastly_nsq/pull/26#discussion_r62533183:
> 
> > @@ -1,8 +1,10 @@
> >  module MessageQueue
> >    class Listener
> > -    def initialize(topic:, channel:)
> > -      @topic = topic
> > -      @channel = channel
> > -    def initialize(topic:, channel:, processor: nil, consumer: nil)
> > -      @topic     = topic
> > -      @channel   = channel
> > -      @processor = processor || DEFAULT_PROCESSOR
> 
> That's true. Is it more readable/intention-revealing in that form? I
> think, long-term, I'd like to see us move away from the "magic constant"
> MessageProcessor in favor of this kind of configuration (maybe with some
> sane defaults), so if we need to establish how we're going to write that
> sort of thing now, maybe it's a good time.
> 
> —
> You are receiving this because you were mentioned.
> https://github.com/fastly/fastly_nsq/pull/26/files/01186203103b8ccd1e5e8ed9fe4914e0bc234ea1#r62533183

## 

Adarsh Pandit
External Consultant
+1 313 454 1515 (m)

README.md
alieander: I like having the Code version of the readme be a little more readable but this probably preferences. I keep linewrap off...

Is it popular here? probably not. Does it matter, not really.
Markdown deals with both in a reasonable manner...

lib/fastly_nsq/message_queue/listener.rb
standingwave, line 6: I prefer the lack of line feeds personally because I like controlling
vertical space. I'll happily follow whatever convention people feel most
strongly about though.

README.md
leklund: I honestly don't have an opinion about linefeeds or line length.

jaw6:
This comment hopes that shiply will run again.

leklund:
:+1:

Issue #27, opened 2016-04-22
Listener spec seems todo nothing.


alieander:
https://github.com/fastly/fastly_nsq/blob/master/spec/lib/fastly_nsq/message_queue/listener_spec.rb#L71-L94

This spec passes with or without the code in https://github.com/fastly/fastly_nsq/blob/master/lib/fastly_nsq/message_queue/listener.rb#L9-L15

alieander:
@adarsh

adarsh:
Agree. This is hard to test because the `SIGINT` will kill the test process as well, so you can't really assert anything that well.

May be worth removing this or asserting that `.terminate` gets sent to the  `MessageQueue::Consumer#connection`.

PR #30, opened 2016-05-10
Remove unnecessary listener spec

adarsh:
# Reason for Change
- This spec does not accurately test the code it is targeting.
- Because the method is testing the behavior of a `SIGTERM`, it actually interferes with the running RSpec process.
- As noted in #27, the test will pass if the relevant code is removed.
# Changes
- Remove the spec in question.
- NOTE: This method is now untested. Since it's the same as another method, but run in a loop, it's probably fine.
# Risks
- One method is now untested, but this feels like a minor risk.
# Checklist

# Recommended Reviewers

@fastly/billing
#27

### DIFF
--- a/spec/lib/fastly_nsq/message_queue/listener_spec.rb
+++ b/spec/lib/fastly_nsq/message_queue/listener_spec.rb
@@ -87,21 +87,4 @@ RSpec.describe MessageQueue::Listener do
       end
     end
   end
-
-  describe '#go' do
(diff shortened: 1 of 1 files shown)


alieander:
:+1:

leklund:
Shiply?

alieander:
Reviewed. Is nice. Would review again. 9/10

PR #29, opened 2016-05-10
Producer#connection -> Producer

jaw6:
Continuing on [from working](https://github.com/fastly/fastly_nsq/pull/26) to make `Consumer#connection` private (in favor of using the actual `MessageQueue::Consumer` instance and delegating) this PR applies the same approach to `Producer#connection`.

@fastly/billing

PR #32, opened 2016-05-10
Persist connection

jaw6:
# Reason for Change

We're seeing NSQ connections leaking after tests.
# List of Changes
- Memoize #connection on Consumer
- Update spec to get around Rspec's mock scoping rules
# Risks

It feels ugly to manipulate test object's instance vars this way. At one point, I had a method like `reset` or something that could do this, but adding a method just for testing also felt ugly. I dunno.
# Recommended Reviewers

@fastly/billing tell me what I'm doing wrong here

### DIFF
--- a/lib/fastly_nsq/message_queue/consumer.rb
+++ b/lib/fastly_nsq/message_queue/consumer.rb
@@ -10,7 +10,7 @@ module MessageQueue
     def_delegator :connection, :size
     def_delegator :connection, :terminate
 
-    def initialize(topic:, channel:, ssl_context: nil, &connector)
+    def initialize(topic:, channel:, ssl_context: nil, connector: nil)
       @topic       = topic
       @channel     = channel
       @ssl_context = SSLContext.new(ssl_context)
--- a/lib/fastly_nsq/message_queue/consumer.rb
+++ b/lib/fastly_nsq/message_queue/consumer.rb
@@ -10,18 +10,21 @@ module MessageQueue
     def_delegator :connection, :size
     def_delegator :connection, :terminate
 
-    def initialize(topic:, channel:, ssl_context: nil)
+    def initialize(topic:, channel:, ssl_context: nil, &connector)
       @topic = topic
       @channel = channel
       @ssl_context = SSLContext.new(ssl_context)
+      @connector = connector || DEFAULT_CONNECTOR
     end
 
     private
 
-    attr_reader :channel, :topic, :ssl_context
+    attr_reader :channel, :connector, :topic, :ssl_context
+
+    DEFAULT_CONNECTOR = ->(params)  { Strategy.for_queue::Consumer.new(params) }
 
     def connection
-      @connection ||= Strategy.for_queue::Consumer.new(params)
+      @connection ||= connector.call(params)
     end
 
     def params
--- a/lib/fastly_nsq/message_queue/consumer.rb
+++ b/lib/fastly_nsq/message_queue/consumer.rb
@@ -21,7 +21,7 @@ module MessageQueue
 
     attr_reader :channel, :connector, :topic, :ssl_context
 
-    DEFAULT_CONNECTOR = ->(params)  { Strategy.for_queue::Consumer.new(params) }
+    DEFAULT_CONNECTOR = ->(params) { Strategy.for_queue::Consumer.new(params) }
 
     def connection
       @connection ||= connector.call(params)
--- a/lib/fastly_nsq/message_queue/consumer.rb
+++ b/lib/fastly_nsq/message_queue/consumer.rb
@@ -11,10 +11,10 @@ module MessageQueue
     def_delegator :connection, :terminate
 
     def initialize(topic:, channel:, ssl_context: nil, &connector)
-      @topic = topic
-      @channel = channel
+      @topic       = topic
+      @channel     = channel
       @ssl_context = SSLContext.new(ssl_context)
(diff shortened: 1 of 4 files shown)


adarsh:
> It feels ugly to manipulate test object's instance vars this way. 

I agree, although (just my two cents) my approach is to avoid using ivars in test. As a Martin Fowler student, it feels like a [Mystery Guest](http://xunitpatterns.com/Obscure%20Test.html), like `before` blocks and `let`s... maybe too aggressively DRY?

More practically, I'm regularly surprised by the contents of a variable with a common name like `user` if it's set outside of my unit test. Perhaps this is just me not being very smart.

FWIW, I actually agree with RSpec on this one (again, my gently-held personal opinion) - I would probably be surprised if something was mock/stubbed outside of my actual test block. 

All of this aside, I ran this branch against my dev env and this appears to fix the connection issue so a big 🍊  from me, regardless of the ivar/doubling stuff.

jaw6:
@adarsh Since we didn't emergency-merge this last night, I spent a little time thinking about a better approach, here & #29. I think you should re-:eyes: ?

lib/fastly_nsq/message_queue/consumer.rb
```diff
@@ -10,18 +10,21 @@ class Consumer
...
     def_delegator :connection, :terminate
 
-    def initialize(topic:, channel:, ssl_context: nil)
-      @topic = topic
-      @channel = channel
+    def initialize(topic:, channel:, ssl_context: nil, &connector)
+      @topic       = topic
+      @channel     = channel
       @ssl_context = SSLContext.new(ssl_context)
+      @connector   = connector || DEFAULT_CONNECTOR
```
adarsh, line 17: Just my two cents, I kind of liked this PR before this change.

This feels now probably more general than we need it to be, although it seems to make testing easier.

Also, and this is not scientific or a well-described perspective by me (sorry), but I haven't seen the constructor take a block and it seems a little funky. I'd expect either the connector to be passed in as an arg or have some other method take a block.

So I'm a -1 on this recent change, but it's like a 3 out of 10 for me.

jaw6, line 17: I'm not sure I understand the objection to constructor-with-block, but since [we have a similar-but-different pattern in play elsewhere](https://github.com/fastly/fastly_nsq/blob/7d70eb82ec50200166b905d64db1924bf99f4a51/lib/fastly_nsq/message_queue/listener.rb#L6), it's certainly easy enough to follow along.

leklund:
:+1: :shipit:

spec/lib/fastly_nsq/message_queue/consumer_spec.rb
```diff
@@ -17,49 +14,47 @@ def fake_consumer
...
+    it 'forwards #pop' do
+      expect(backend).to receive(:pop)
       consumer.pop
-      expect(@fake_consumer).to have_received(:pop)
     end
 
-    it 'forwards #size to Nsq::Consumer' do
+    it 'forwards #pop_without_blocking' do
+      expect(backend).to receive(:pop_without_blocking)
+      consumer.pop_without_blocking
```
adarsh, line 29: You can use shoulda-matchers here if you like:

``` ruby
it { is_expected.to delegate_method(:pop_without_blocking).to(:backend) }
```

http://www.rubydoc.info/github/thoughtbot/shoulda-matchers/Shoulda%2FMatchers%2FIndependent%3Adelegate_method

adarsh, line 29: Oops, this already merged. Please disregard.

jaw6, line 29: No worries, always good to have another TIL thing.

That said ... I dunno about these, I have mixed feelings about "specialized" matchers. I think it makes sense when they come from the "app domain" (I dunno... "expect to have_connected_to_backend" maybe?) but, for me, `shoulda_matchers` feel like they come from the "rspec domain" (or, best case scenario, the "I am writing ruby domain"), which doesn't exactly sit right with me. `¯\_(ツ)_/¯`

adarsh, line 29: No worries - I take the point for sure :)

On Thu, May 12, 2016 at 12:33 PM, Joshua Wehner notifications@github.com
wrote:

> In spec/lib/fastly_nsq/message_queue/consumer_spec.rb
> https://github.com/fastly/fastly_nsq/pull/32#discussion_r63082574:
> 
> > ```
> >  end
> > ```
> > -    it 'forwards #size to Nsq::Consumer' do
> > -    it 'forwards #pop_without_blocking' do
> > -      expect(backend).to receive(:pop_without_blocking)
> > -      consumer.pop_without_blocking
> 
> No worries, always good to have another TIL thing.
> 
> That said ... I dunno about these, I have mixed feelings about
> "specialized" matchers. I think it makes sense when they come from the "app
> domain" (I dunno... "expect to have_connected_to_backend" maybe?) but, for
> me, shoulda_matchers feel like they come from the "rspec domain" (or,
> best case scenario, the "I am writing ruby domain"), which doesn't exactly
> sit right with me. ¯_(ツ)_/¯
> 
> —
> You are receiving this because you were mentioned.
> https://github.com/fastly/fastly_nsq/pull/32/files/eece8723e93aa9b48166019dfebfea510f185f02#r63082574

## 

Adarsh Pandit
External Consultant
+1 313 454 1515 (m)

PR #33, opened 2016-05-11
Strategy

jaw6:
# Reason for Change

Previously, the spec_helper set us up to run each test 7-8 times via:

``` ruby
  def use_fake_connection
    MessageQueue::Strategy::TRUTHY_VALUES.each do |yes|
      allow(ENV).to receive(:[]).with('FAKE_QUEUE').and_return(yes)
      yield
    end
  end
```
# List of Changes
- Reduce that back down to once
- Re-namespace & refactor `Strategy` (via "campsite rule")
# Risks

There are likely to be merge conflicts later with #32 & #29 

@fastly/billing

### DIFF
--- a/lib/fastly_nsq/message_queue/strategy.rb
+++ b/lib/fastly_nsq/message_queue/strategy.rb
@@ -8,5 +8,8 @@ class Strategy
       message = "You must set ENV['FAKE_QUEUE'] to either true or false"
       raise InvalidParameterError, message
     end
+
+  FALSY_VALUES  = [false, 0, '0', 'false', 'FALSE', 'off', 'OFF', nil].freeze
+  TRUTHY_VALUES = [true, 1, '1', 'true', 'TRUE', 'on', 'ON'].freeze
   end
 end
--- a/lib/fastly_nsq/message_queue/strategy.rb
+++ b/lib/fastly_nsq/message_queue/strategy.rb
@@ -1,15 +1,34 @@
-class Strategy
-  def self.for_queue
-    if MessageQueue::FALSY_VALUES.include?(ENV['FAKE_QUEUE'])
-      Nsq
-    elsif MessageQueue::TRUTHY_VALUES.include?(ENV['FAKE_QUEUE'])
-      FakeMessageQueue
-    else
-      message = "You must set ENV['FAKE_QUEUE'] to either true or false"
-      raise InvalidParameterError, message
-    end
+module MessageQueue::Strategy
+  module_function
+
+  def for_queue
+    real_queue || fake_queue || error
+  end
+
+  private_class_method
+
+  ERR_MESSAGE = "You must set ENV['FAKE_QUEUE'] to either true or false".freeze
+
+  def error
+    raise InvalidParameterError, ERR_MESSAGE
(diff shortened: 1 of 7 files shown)


lib/fastly_nsq/message_queue/strategy.rb
```diff
@@ -1,12 +1,34 @@
...
-      raise InvalidParameterError, message
-    end
+module MessageQueue::Strategy
+  module_function
+
+  def for_queue
+    real_queue || fake_queue || error
+  end
+
+  private_class_method
```
adarsh, line 8: [Based on the source](http://ruby-doc.org/core-2.3.1/Module.html#method-i-private_class_method), this looks like it takes a symbol as an argument, like

``` ruby
class Foo
  private_class_method(:method_name)
  # ...
```

Is there a missing argument or am I misreading the usage?

lib/fastly_nsq/message_queue/strategy.rb
```diff
@@ -1,12 +1,34 @@
...
-    if MessageQueue::FALSY_VALUES.include?(ENV['FAKE_QUEUE'])
-      Nsq
-    elsif MessageQueue::TRUTHY_VALUES.include?(ENV['FAKE_QUEUE'])
-      FakeMessageQueue
-    else
-      message = "You must set ENV['FAKE_QUEUE'] to either true or false"
-      raise InvalidParameterError, message
-    end
+module MessageQueue::Strategy
+  module_function
```
adarsh, line 2: Same here: does this take a symbol as an argument?

http://ruby-doc.org/core-2.3.1/Module.html#method-i-module_function

lib/fastly_nsq/message_queue/strategy.rb
```diff
@@ -1,12 +1,34 @@
...
+
+  def for_queue
+    real_queue || fake_queue || error
```
adarsh, line 5: I like this ✨

adarsh:
Some minor questions, otherwise LGTM.

lib/fastly_nsq/message_queue/strategy.rb
alieander, line 2: From that document link:

> If used with no arguments, subsequently defined methods become module functions.

adarsh, line 2: Ha! I can't read my own links... 📚

lib/fastly_nsq/message_queue/strategy.rb
alieander, line 8: I assume this acts like the above `module_function` but it is less clear...

jaw6, line 8: It's kind of Ruby minutiae: `private` overrides `module_function`, they don't "synergize". If you want a private, class-level method (or module-level method, as this case would have it), `private_class_method` is the thing. Personally, I think there's [an argument to made](http://blog.codeclimate.com/blog/2012/11/14/why-ruby-class-methods-resist-refactoring/) that these class-(module-)level methods aren't great, but I think it's the best scenario here. (aka "the tradeoffs are worth it")

PR #34, opened 2016-05-23
Update TravisCI > Slack integration key

adarsh:
# Reason for Change
- We had previously committed a Slack API key for Travis notifications to this public repository.
- This is a security risk so the key has been invalidated.
- We need a new secure key instead to maintain notifications.
# Changes
- Invalidate the old integration key in Slack.
- Use the travis CLI tool to [encrypt the existing key](https://docs.travis-ci.com/user/notifications/#Slack-notifications) which we weren't using yet.

https://jira.corp.fastly.com/browse/SEC-1822

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ script:
 notifications:
   slack:
     rooms:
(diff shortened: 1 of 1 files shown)


jaw6:
:+1: :+1:

PR #36, opened 2016-06-06
Remove broadcast Address

alieander:
We are not using it as it defualts to the correct thing everywhere we
are using it


jaw6:
LGTM

PR #37, opened 2016-06-23
Consistent use of namespace & injectable dependencies

jaw6:
Making a proper PR out of this, maybe a bit past due.
### Goal

Ruby gems/libraries are generally expected to package code into a singular namespace. Prior to this PR, this gem violates this principles in two ways:
- _Some_ classes are part of `MessageQueue` module; others are part of `FastlyNsq`; others (`FakeMessageQueue`) are not in any namespace at all.
- The gem requires the existence of a `MessageProcessor` class, which responds to a `.topics` method, to configure `Listener` to listen to the client's topics.
### Benefits
- Clients can use this gem without any danger of collision -- eg, there's no problems if there's already a `MessageProcessor` or `MessageQueue` in client code.
- Cleaner injection of dependencies in `Listener` (goodbye, `DEFAULT_MESSAGE_PROCESSOR`)

There's some additional discussion in [this comment](https://github.com/fastly/fastly_nsq/pull/37#issuecomment-229236604)
### Changes
- Organize all of this gem's code into one module, `FastlyNsq`
- Move files into a folder structure matching this namespace, per gem convention
- Remove `MessageProcessor` dependency, require clients to configure `topics` in the listener `RakeTask`
### Impact

Clients using the `RakeTask` helper will need to update after this change. Instead of calling `MessageProcessor.topics`, the `RakeTask` now requires a topics configuration, something like:

``` ruby
FastlyNsq::RakeTask.new(:listen_task) do |task|
  task.channel = 'some_channel'
  task.topics  = topics
end
```

Here, `topics` should be a Hash (or Array of tuples), where the keys are NSQ topics and the values respond to `.process(message_body, topic)`

I have begun work adapting one of our client projects to this and will take responsibility for updating the others.

/cc @fastly/billing


alieander:
The first commit(73e6b0d392d765fb5e5867199296ad2a813d60d2) alone is big enough to be it's own PR...

spec/lib/fastly_nsq/consumer_spec.rb
```diff
@@ -6,14 +6,6 @@
...
   let(:backend)  { double 'Consumer' }
 
-  describe 'when the ENV is set incorrectly' do
-    it 'raises with a helpful error' do
-      allow(ENV).to receive(:[]).with('FAKE_QUEUE').and_return('taco')
-
-      expect { consumer.terminate }.to raise_error(InvalidParameterError)
-    end
-  end
-
```
alieander: I assume you found these tests unhelpful?

lib/fastly_nsq/listener.rb
```diff
@@ -1,49 +1,45 @@
...
-        process_one_message
-      end
-    end
+        next_message do |message|
+          processor.process(message.body, topic)
+        end
 
-    def process_next_message
-      process_one_message
+        break if limit
```
alieander: `run_once`?

lib/fastly_nsq/rake_task.rb
```diff
@@ -15,68 +16,50 @@ def initialize(*args, &task_block)
             yield(*[self, task_args].slice(0, task_block.arity))
           end
 
-          if task_args[:channel]
-            @channel = task_args[:channel]
-          end
+          @channel  ||= require_arg :channel, task_args 
+          @topics   ||= require_arg :topics, task_args
+          @listener ||= task_args[:listener]
```
alieander: Passing topics and channels into the listener is where we started.
This moves the addition of new topics (ATM) to listen to into a cookbook change.

I assume this is an intermediary step to something more like a CLI where you inform it of configuration via flags? and do something akin to what the `sidekiq` cli does by loading a require path?

lib/fastly_nsq/version.rb
```diff
@@ -1,3 +1,3 @@
 module FastlyNsq
-  VERSION = '0.6.0'.freeze
+  VERSION = '0.7.0'.freeze
```
alieander, line 2: If this is breaking and non-backwards compatible, and changing the name you must use is exactly that, then the correct thing todo here would be to bump the `MAJOR` not the `MINOR`...

jaw6:
> The first commit alone is big enough to be it's own PR...

@alieander it very nearly was, but I had trouble justifying pushing a big breaking change like that without more justification. 

For what it's worth, my sense is that it a) needs to happen b) needs to happen pre 1.0 c) will break things.

Are you requesting that it be broken out into its own PR now? (This is still WIP, but your comment is unclear.)

lib/fastly_nsq/version.rb
jaw6, line 2: @alieander I think you're trying to apply a rule from [SemVer](http://semver.org/). At least, I'm assuming -- I don't see where we've explicitly committed the project to using SemVer (there are differences among versions, things that aren't quite SemVer but look like it, etc).

My understanding of SemVer is that the rule you seem to be trying to apply does not apply (yet):

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

And

> Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

It _could_ be that you are suggesting we're ready for v1.0, which I would rather strongly disagree with.

Can you elaborate a bit on what you want here?

jaw6:
> I assume you found these tests unhelpful?

@alieander the tests you're referencing are tests in `Consumer` and `Producer` that specify that `Strategy` raises errors if the `FAKE_QUEUE` Env var is an unexpected value. I can see how this would be valuable when developing `Consumer` and `Producer` initially, and evolving into the `Strategy` class we have now, but as we now have a class who owns this responsibility, and there's no `Consumer`\- or `Producer`-specific behavior under test (we're _not_ specifying how those classes handle the exceptions, for example), it doesn't make sense to perpetuate them. They served their purpose and can go.

lib/fastly_nsq/listener.rb
jaw6: `run_once` is a better name than `limit` -> fixed in 5a6c95b

lib/fastly_nsq/rake_task.rb
jaw6: > This moves the addition of new topics (ATM) to listen to into a cookbook change

@alieander I'm missing something, but I don't see that happening. I've got a Proof of Concept / Demo put together in one of our apps, but it doesn't require cookbook changes, as far as I can tell. Elaborate if you can on what you think I'm missing?

lib/fastly_nsq/version.rb
alieander, line 2: I appear to have not read far enough down.

alieander:
> Are you requesting that it be broken out into its own PR now? (This is still WIP, but your comment is unclear.)

Sorry for the unclear comment. It was just a comment. I noted no real action that needed to be taken.

alieander:
> the tests you're referencing are tests in Consumer and Producer that specify that Strategy raises errors if the FAKE_QUEUE Env var is an unexpected value. I can see how this would be valuable when developing Consumer and Producer initially, and evolving into the Strategy class we have now, but as we now have a class who owns this responsibility, and there's no Consumer- or Producer-specific behavior under test (we're not specifying how those classes handle the exceptions, for example), it doesn't make sense to perpetuate them. They served their purpose and can go.

This is useful context that the commit message lacked. Seeing tests disappear without being replaced and without being commented on is a light flag for me... so I asked to be sure.

adarsh:
> Remove references to MessageProcessor
>  ...
> For what it's worth, my sense is that it a) needs to happen b) needs to happen pre 1.0 c) will break things.

There isn't much documentation in this PR as to the motivation for this change, what the goals are, or why it "needs to happen". Can you say more here to help understanding (particularly given these changes are breaking in nature)?

jaw6:
> There isn't much documentation in this PR as to the motivation for this change, what the goals are, or why it "needs to happen". Can you say more here to help understanding (particularly given these changes are breaking in nature)?

Sure, though, fwiw, I'm pretty sure we've discussed all the reasons IRL, but documenting reasons is never a bad idea: 
- `MessageProcessor` is primarily (maybe _only_) used to map NSQ Topics to their respective "processors".
- `MessageProcessor` was introduced in #22. At the [time](https://github.com/fastly/fastly_nsq/blob/d93e40fce182e051d4fd92240aff42c9b2a7b7ca/lib/fastly_nsq/rake_task.rb#L16), the rake task helper required a `topic`, to which it would configure the listener to listen to. The change introduced `MessageProcessor` and `.topics` as a way to listen to more than one topic at once. Despite being a breaking change at the time (although #22 does a good job calling attention to this breakage), the only discussion of motivation/goals given at the time appears to have been "Don't pass topics into the Raketask, specify in `MessageProcessor.topics`." I don't see any further discussion in [the PR](https://github.com/fastly/fastly_nsq/pull/22).
- It's not, strictly speaking, a _requirement_ to pass `topics` this way. It's not, strictly speaking, used at all really (other than shenanigans with `SampleMessageProcessor` in tests, it's barely used in the gem at all), it's just the means we've chosen to convey a particular bit of configuration from the client using the gem into the gem itself, so that `Listener` can call back out to the right client code.
- In a [world without dependency injection](https://github.com/fastly/fastly_nsq/blob/27b2acb99945fa7114447414540cfd06f31b2790/lib/fastly_nsq/message_queue/listener.rb#L33) discovering that `Listener` depends on a not-declared-in-this-gem constant in the middle of a private method is a violation of, at _least_, the Principle of Least Astonishment (sure; it's called out as a requirement in the Readme, but a Readme is, at best, a page(s) long code comment).
- In the course of fixing our abstractly and practically leaking `Nsq::Producer` and `Nsq::Consumer` objects, I [introduced an injectable dependency for `MessageProcessor`](https://github.com/fastly/fastly_nsq/blob/d1ebc2554f579d8b04e251e971651493ac98c1f2/lib/fastly_nsq/message_queue/listener.rb#L32), where the dependency could be injected as a Proc or lambda. As a half-assed form of DI, this works well enough, but it also adds a layer of indirection that's more confusing than clarifying. Recent discussions with the team lead to the conclusion that we should "complete" injectable dependency refactorings like this, rather than leave them in a half-way-there state.
- All of the confusion and complication is entirely unnecessary -- the worker/topic mapping is still _being configured_ (this is _not_ "convention > configuration"), we've just moved the topic/worker mapping off a distance from where it's required for reasons that aren't at all clear, and codified that mapping in the form of a dependency on a specifically-named magical, global variable.
- When a client uses the listener to observe topics and process their messages, they primarily do so via the gem's included [rake task helper](https://github.com/fastly/fastly_nsq/blob/d1ebc2554f579d8b04e251e971651493ac98c1f2/lib/fastly_nsq/rake_task.rb#L19) which also includes a configurable `channel` (in fact, setting the channel in the Rake task is _required_, not optional) so it's not as if we're opposed to having configuration happen in the Rake task, we're just not using it for topic/worker mappings, for reasons never made clear.
- If there's already a `MessageProcessor` in a downstream client, there's no option for them to use this gem. This is impolitic of us -- we have namespaces in Ruby and should use them -- and, in a way, using a magical, global variable, just so that the `Listener` can get a topic/worker mapping. 

Tl;dr: [Global Variables Are (still) Bad](http://c2.com/cgi/wiki?GlobalVariablesAreBad) and this one isn't remotely necessary, this set of changes makes it go away.

README.md
```diff
@@ -167,7 +161,7 @@ or using passed in values:
 ```ruby
 require 'fastly_nsq/rake_task'
 
-MessageQueue::RakeTask.new(:listen_task, [:channel])
+FastlyNsq::RakeTask.new(:listen_task, [:channel])
```
alieander: I am not sure this is a sound example any longer as topic is required.

alieander:
the `example/Rakefile` needs to be changed. it still references `MessageProcessor` and all of the tasks describe rely on its existence rather than passing a topics hash.

README.md
jaw6: :+1: fixed in b958877

lib/fastly_nsq/rake_task.rb
```diff
@@ -15,68 +16,54 @@ def initialize(*args, &task_block)
...
-          guard_missing_channel
-          run_tasks
+          do_stuff
         end
       end
     end
 
     private
 
+    def do_stuff
```
alieander: Can this get a more meaningful name?

jaw6:
> the example/Rakefile needs to be changed

Fixed in 0d30f8d

lib/fastly_nsq/rake_task.rb
jaw6: :+1: good catch; `listen_to_configured_topics` in aec9e45

alieander:
👍

jaw6:
@alieander After our conversation about logging messages inside the gem, and further developments in our main client app, I made the changes here necessary to support both: logging now happens inside the gem, but also there is a `preprocessor` hook that allows clients to perform additional "processing" before `processor.process`. In this case, all listener/topic/processor instances will use the same `preprocessor` but that suits our use-case pretty well.

I'm also opening and/or updating PRs on the client applications bumping them to this version of the gem, with associated changes in namespacing. 

Given the changes in the commits above, this might need one more :eyes:.

/cc @standingwave, who paired with me a bit on some of the last batch of changes.

lib/fastly_nsq/listener.rb
```diff
@@ -4,10 +4,11 @@ def self.listen_to(**args)
       new(**args).go
     end
 
-    def initialize(topic:, processor:, channel: nil, consumer: nil)
-      @topic     = topic
-      @processor = processor
-      @consumer  = consumer || FastlyNsq::Consumer.new(topic: topic, channel: channel)
+    def initialize(topic:, processor:, channel: nil, consumer: nil, preprocessing: nil)
+      @topic         = topic
+      @preprocessing = preprocessing
```
leklund: Would be nice to see a note about the preprocessor in the README or the examples

lib/fastly_nsq/listener.rb
```diff
@@ -16,6 +17,7 @@ def go(run_once: false)
 
       loop do
         next_message do |message|
+          preprocessing.call(message.body) if preprocessing
```
leklund: Was going to say we might want to check `if preprocessing.is_a?(Proc)` but as long as it's documented I don't think that's necessary.

lib/fastly_nsq/listener.rb
```diff
@@ -0,0 +1,59 @@
+require 'fastly_nsq/message'
+
+module FastlyNsq
+  class Listener
+    def self.listen_to(**args)
+      new(**args).go
+    end
+
+    def initialize(topic:, processor:, channel: nil, consumer: nil, **options)
```
leklund, line 9: can `channel` ever be nil? should it be required?

leklund:
I'd like to see the README updated with the new `logger` and `preprocessor` options but other than that I think this looks good.

lib/fastly_nsq/listener.rb
jaw6, line 9: It's hard to see here with the rename, but keep in mind this has [not really changed relative to current master](https://github.com/fastly/fastly_nsq/blob/f0da289828b02c0fd9683ddb1832257e58b3b7aa/lib/fastly_nsq/message_queue/listener.rb).

Originally, `Listener` required `topic` & `channel` and managed `consumer` internally. (`topic` is still passed to `processor`, but currently all of our clients ignore it, so it, too, could be made optional.) 

Personally, I'd eventually like to see this drop `channel` _and_ `topic`, and initialize `Listener` with a `Consumer`. The main challenge here is that `Listener`'s specs indicate that it sort of wants to wrap & manage `Consumer` (lazily instantiating and eventually terminating). 

I'd like to move this up into something like `FastlyNsq.listen_to`, let that method coordinate `Consumer` and simplify `Listener` a great deal) but I felt this PR was already sprawling in scope, and this change wasn't necessary even in the (currently enlarged) scope.

lib/fastly_nsq/listener.rb
```diff
@@ -0,0 +1,59 @@
...
+      @topic        = topic
+      @processor    = processor
+      @consumer     = consumer || FastlyNsq::Consumer.new(topic: topic, channel: channel)
+      @logger       = options.fetch :logger, FastlyNsq.logger
+      @preprocessor = options[:preprocessor]
+    end
+
+    def go(run_once: false)
```
alieander, line 17: to be honest I wish this was two methods... `run` and `run_once`

jaw6, line 17: Discussed this IRL: refactoring makes this method more clear; `go` could just be called `run` with a companion method `run_once`, even if the `run_once` method were only used in tests, it's a small difference compared to passing in as a parameter. For now, leaving with one `go` method (and parameter) but could be something to refactor in a future PR.

leklund:
This gets a 👍 🎉 👯  from me

PR #38, opened 2016-06-26
call Thread#join only on threads we created

alieander:
# Reason for Change

We where using `Thread.list` which could likely contain things we did not
start. This switches to saving the threads we create in an array and
calling `Thread#join` on only the ones we have started.

We could also select only non-dead threads from this list by doing

```
listener_threads.select(&:status).map(&:join)
```

but by including threads that have already died, we allow them to
express whatever exception might have been raised back to the main
thread effectively halting the main Thread as they are left un-handled
there.`Thread.list` would have not returned "dead" threads here as we
where using it, so it was possible for a Thread to fail immediately and
have `join` never be called on it making it's failure silent.
# List of Changes
- alter rake task to save list of created listeners rather than use `Thread.list`
# Recommended Reviewers

@fastly/billing

### DIFF
--- a/lib/fastly_nsq/rake_task.rb
+++ b/lib/fastly_nsq/rake_task.rb
@@ -44,7 +44,11 @@ module MessageQueue
         end
       end
 
-      listener_threads.map(&:join)
+      while listener_threads.any?(&:status) do
+        listener_threads.each do |thread|
+          thread.join(1)
+        end
+      end
     end
 
     def guard_missing_channel
--- a/lib/fastly_nsq/rake_task.rb
+++ b/lib/fastly_nsq/rake_task.rb
@@ -44,10 +44,11 @@ module MessageQueue
         end
       end
 
-      while listener_threads.any?(&:status)
+      loop do
         listener_threads.each do |thread|
           thread.join(1)
         end
+        break unless listener_threads.any?(&:status)
       end
     end
 
--- a/lib/fastly_nsq/rake_task.rb
+++ b/lib/fastly_nsq/rake_task.rb
@@ -44,7 +44,7 @@ module MessageQueue
         end
       end
 
-      while listener_threads.any?(&:status) do
+      while listener_threads.any?(&:status)
         listener_threads.each do |thread|
           thread.join(1)
         end
--- a/lib/fastly_nsq/rake_task.rb
+++ b/lib/fastly_nsq/rake_task.rb
@@ -34,19 +34,17 @@ module MessageQueue
     end
 
     def run_tasks
+      listener_threads = []
+
       topics.each do |topic|
-        Thread.new do
+        listener_threads << Thread.new do
           wrap_helpful_output(topic) do
             MessageQueue::Listener.new(topic: topic, channel: channel).go
           end
         end
       end
 
-      non_main_threads.map(&:join)
-    end
-
-    def non_main_threads
-      (Thread.list - [Thread.main])
+      listener_threads.map(&:join)
     end
 
     def guard_missing_channel
--- a/lib/fastly_nsq/rake_task.rb
+++ b/lib/fastly_nsq/rake_task.rb
@@ -37,19 +37,16 @@ module MessageQueue
       listener_threads = []
 
       topics.each do |topic|
-        listener_threads << Thread.new do
+        thread = Thread.new do
           wrap_helpful_output(topic) do
             MessageQueue::Listener.new(topic: topic, channel: channel).go
           end
         end
+        thread.abort_on_exception = true
+        listener_threads << thread
       end
 
-      loop do
-        listener_threads.each do |thread|
-          thread.join(1)
-        end
-        break unless listener_threads.any?(&:status)
-      end
+      listener_threads.map(&:join)
     end
 
     def guard_missing_channel


standingwave:
> but by including threads that have already died, we allow them to express whatever exception might have been raised back to the main thread effectively halting the main Thread as they are left un-handled there.

Test to confirm this behavior?

leklund:
It's not a test as such, but here is a ruby block that confirms this behavior (I was curious):

```
## Thread.list
Thread.new do
  raise 'an error'
end

sleep 1

t = Thread.list  - [Thread.main]
t.map(&:join)

puts 'This prints and does not raise'


## array of threads
lt = []
lt << Thread.new do
  raise 'an error'
end

sleep 1

lt.map(&:join)

puts 'this never prints because the thread raises.'
```

lib/fastly_nsq/rake_task.rb
```diff
@@ -44,7 +44,11 @@ def run_tasks
         end
       end
 
-      listener_threads.map(&:join)
+      while listener_threads.any?(&:status) do
+        listener_threads.each do |thread|
+          thread.join(1)
+        end
+      end
```
leklund: Is checking status here necessary? As long as any one thread has a truthy `status` it will still try to join all the threads we are tracking, even the threads that have exited. I guess it is so that you can try to join every second until all threads are joined. Super tiny race condition where all threads exit before any it reaches this block but that seems unlikely/impossible.

leklund:
added a comment where I did some thinking out loud. This looks reasonable to me.

lib/fastly_nsq/rake_task.rb
```diff
@@ -34,19 +34,21 @@ def add_rake_task_description_if_one_needed
...
       end
 
-      non_main_threads.map(&:join)
-    end
-
-    def non_main_threads
-      (Thread.list - [Thread.main])
+      while listener_threads.any?(&:status)
+        listener_threads.each do |thread|
+          thread.join(1)
```
jaw6: `threads.any?` and `thread.each` both loop through all threads, which makes me think that the outer/inner loop split here doesn't save a ton. What about:

```
listener_threads.each do |thread|
  thread.join(1) if thread.status
end
```

leklund: I had the same idea, but this will only loop each thread once. And should a thread raise before it becomes the subject of the loop it will be silently discarded. It also won't wait for threads to join because it will only wait one second and then return nil if the thread has not exited.

@alieander: I assume you used the `while` and the `join(1)` so that if one thread raised an error it will be caught here? What about using `abort_on_exception=true` so that when a thread raises we'll just raise and exit?

lib/fastly_nsq/rake_task.rb
alieander: With the addition of a time (one second here), `Thread#join` will return to the main thread after that time regardless of what happens in the sub thread. The inner `.each` will loop over all threads once, and then be done taking at most `threads.count` seconds.
The check of `Thread#status` keeps the while loop looping while there are still threads that are available to be `#join`'ed

The case where `.any?` is reached is a good point. I can change this to a `begin` `while` loop so that the join's are executed at least once.

lib/fastly_nsq/rake_task.rb
alieander: Double looping here keeps us looping until all threads have finished in some way. Before we blocked on the first thread until it exited... and then on the second after that and so on. This got a process that eventually looped over each thread until all had finished. If we want to allow each one to raise errors (so we can do exception_handling and restarts maybe), we can limit, or we can set `abort_on_exception=true`. The later though this is not guaranteed to raise when you would expect back to the main thread (reading on the internet this weekend gave me this impression) so we would have to know more about the nuances there. You can set this with the `$DEBUG` flag which makes me think it was more for debugging than aborting early?...

This should likely be more verbose (include exception_handling etc.) but I did not want to make to great a change 1: over the weekend, 2: while @jaw6's large PR is still open.

alieander: Changed it to use `Thread.abort_on_exception=` as that seems less likely to confuse.

lib/fastly_nsq/rake_task.rb
```diff
@@ -37,19 +37,16 @@ def run_tasks
       listener_threads = []
 
       topics.each do |topic|
-        listener_threads << Thread.new do
+        thread = Thread.new do
           wrap_helpful_output(topic) do
             MessageQueue::Listener.new(topic: topic, channel: channel).go
           end
         end
+        thread.abort_on_exception = true
```
jaw6, line 45: I _think_ I like this better than the loop (`unless status` made my brain a little hurty), but I'm curious: why not `Thread.abort_on_exception` ?

alieander, line 45: `Thread.abort_on_exception` sets the global state for all threads. As this is a Rake task in a Gem, it seems rather rude to enforce this on all Threads made post this call when it is possible todo at the instance level.

These threads will still cause all others to exit if they raise an exception, but other threads made can act as they choose.

jaw6, line 45: Oh, that makes a lot of sense. :+1:

PR #40, opened 2016-10-26
Allow the use of multiple lookups

alieander:
This changes the way we deal with the ENV variable to allow for single
lookup addresses or lists of addresses for nsq lookupd.


leklund:
Storing the array of values in the string ENV var isn't my favorite but it's the best option since we are using dotenv. :+1:

PR #41, opened 2016-11-01
Ensure Thread::Queue gets closed

jaw6:
@fastly/billing especially @alieander, @standingwave + @set5think 

producer.write returns a Thread::Queue which persists until garbage collection, potentially increasing globally open threads to the maximum, which can be problematic.

In applications that persist a singleton producer for most writes, this likely isn't a problem; however, in applications that instantiate individual 'one-off' producers (eg, for a rake task), this can lead to resource starvation.

Note that this is effectively opt-in: downstream apps would need to switch from `producer.write` to `messenger.deliver` to get the auto-closing Thread::Queue.


set5think:
This is a pretty nice implementation, in that it closes all the threads and the queue without explicitly needing to.

I still think we, at the least, need to alert nsq-ruby about the issue and have it fixed there, because it's way too easy to use that interface and not have someone notice it because everything looks right about it.

jaw6:
> I still think we, at the least, need to alert nsq-ruby about the issue

:+1: I agree. `nsq-ruby`'s implementation is a little complex; it's non-trivial to resolve at that level. I thought I'd start out fixing our issues before they spread to our other services. I should also open an issue with `nsq-ruby` (unless that's part of @alieander's plan)

alieander:
I have not opened issues there in the past :P
Only PR's... 

This seems like it will do the trick for us. Just need to use it in all three places.

jaw6:
Do we still need this after https://github.com/wistia/nsq-ruby/pull/32 ?

alieander:
Likely not...

jaw6:
Superseded by #43

PR #42, opened 2016-11-23
Upgrade to nsq-ruby version 2.0.3

alieander:
This upgrade gains some performance fixes that we need to avoid Thread
leakiness but comes with a slightly newer verbage for tls/ssl options
passed in.

Changes
-------
* upgrade to nsq-ruby 2.0.3 from 1.6.0
* change SSLContext class to be called TlsOptions to new gems wants
* add support for passing in a verify_mode


lib/fastly_nsq/tls_options.rb
```diff
@@ -1,15 +1,18 @@
 module FastlyNsq
-  class SSLContext
+  class TlsOptions
     def initialize(context = nil)
       @context = context || {}
     end
 
     def to_h
       merge_contexts
       if empty_context?
```
leklund: I didn't think `empty_context?` would work anymore since you delete all the nil key's in `merge_context`. But I was wrong:

```
2.3.1 :001 > {}.all?{|k,v| v.nil?}
 => true
```

still, this might be better off as

```
@context.empty?
```

lib/fastly_nsq/consumer.rb
```diff
@@ -11,13 +11,13 @@ class Consumer
     def initialize(topic:, channel:, ssl_context: nil, connector: nil)
       @topic       = topic
       @channel     = channel
-      @ssl_context = SSLContext.new(ssl_context)
+      @tls_options = TlsOptions.new(ssl_context).to_h
```
leklund: Seems a little odd that you change `SslContext` to `TlsOptions` but left `ssl_context` in the method signature. I know that means less downstream changes but it definitely is a bit confusing.

I also think the `to_h` should be inside the merge, but I don't really have a great reason why.

lib/fastly_nsq/tls_options.rb
alieander: This is a good point. I'll change this .

lib/fastly_nsq/consumer.rb
alieander: I don't think we use the passed params anywhere ATM... we lean on the `ENV` variables as defaults so changing that shouldn't make downstream work much harder at all...

leklund:
I like it. Let's :shipit:

PR #43, opened 2016-12-20
Reuse connections

leklund:
Reason for Change
===================
* When creating and terminating many `Nsq::Producer`s, there seems to be a memory leak (https://github.com/wistia/nsq-ruby/issues/34)

This attempts to persist the Producer objects for re-use.

List of Changes
===============
* Adds Fastly::Messenger class inspired by #41 (I used @jaw6's work as a starting point)
* Persists the producers in a hash and adds convince methods for retrieving and terminating.

Risks
=====
* My informal tests showed that using the same producer was thread safe with sidekiq, but there may be unintended consequences I haven't considered.

Recommended Reviewers
=====================
@fastly/billing


jaw6:
LGTM :+1:

PR #44, opened 2017-01-04
update nsq-ruby dependency

leklund:
* Update nsq-ruby dependency to include fixes from https://github.com/wistia/nsq-ruby/issues/34


standingwave:
Awesome! Nice work pushing for a fix upstream.

PR #45, opened 2017-01-10
we should produce to a lookup as well

alieander:
This swaps out the nsqd env variable for the nsqlookupd
variable. This may... or may not help things. The conversation with the lookups does not happen every time a message is being worked with, only on first connection and if something goes wrong.
Considering the way we use producers as single purpose things, this might have to talk to lookup every time we want to write a message which could introduce some latency issues.

The producer should now get, from the lookup, the nsqd that is ready for its
message.

HYD-1505

### DIFF
--- a/spec/lib/fastly_nsq/fake_backend_spec.rb
+++ b/spec/lib/fastly_nsq/fake_backend_spec.rb
@@ -57,7 +57,7 @@ RSpec.describe FastlyNsq::FakeBackend::Message do
       topic = 'death_star'
       content = 'hello'
       producer = FastlyNsq::FakeBackend::Producer.new(
-        nsqd: ENV.fetch('NSQD_TCP_ADDRESS'),
+        nsqlookupd: ENV.fetch('NSQD_TCP_ADDRESS'),
         topic: topic,
       )
       producer.write(content)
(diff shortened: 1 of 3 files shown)


spec/lib/fastly_nsq/fake_backend_spec.rb
```diff
@@ -57,7 +57,7 @@
       topic = 'death_star'
       content = 'hello'
       producer = FastlyNsq::FakeBackend::Producer.new(
-        nsqd: ENV.fetch('NSQD_TCP_ADDRESS'),
+        nsqlookupd: ENV.fetch('NSQD_TCP_ADDRESS'),
```
leklund: shouldn't this be `fetch`ing the `NSQLOOKUPD_HTTP_ADDRESS`?

leklund:
Because of the changes in #43 we are persisting producers so discovery overhead on initialization shouldn't' be a problem.

This will do regular discovery for nsqd's with a default interval of 60 seconds: https://github.com/wistia/nsq-ruby/blob/master/lib/nsq/producer.rb#L18-L21

I think we might want to up that interval to 300 or even 1500 but I'm not sure it matters that much.

spec/lib/fastly_nsq/fake_backend_spec.rb
alieander: Oops!

alieander:
we can leave it at 60 for now and see how it does in the future

PR #47, opened 2017-01-19
Wait for Producer to connect.

alieander:
When working with a `Producer` over `nsqlookupd`'s, there is a very short time where the Thread that attempts to establish connections might not have found one yet. In the case we have before this PR, we are always going to run into this problem with the first `.write` we make as we lazily make the `Producer` and then call `.write` straight away.

This changes our initializer to be less lazy AND has it wait until the `Producer` believes it is "connected".


leklund:
:+1:

~NO~
SLEEP
TIL
~BROOKLYN~CONNECTED?

alieander:
lol

PR #49, opened 2017-05-16
Setup a possible path to daemonize listeners

alieander:
This is a rather direct copy of Sidekiq's CLI made to work with our existing
listeners. It still creates one listener per topic, but does so in
a bit more of a managed way.

An example of how this would be used:
```bash
./bin/fastly_nsq -r ./example_config_class.rb -L ./tester.log -P ./piddy.pid -v -d
```
(I have committed `example_config_class.rb` to support review)

What is going on?
------------------

First we add some classes:
* `CLI`: a class that parse opts, sets up some state, helps daemonize if needed, kicks things off
* `Launcher`: very light wrapper of thread manager and heartbeat
* `Manager`: tracks running listeners, restarts them as needed
* `Listener::Config`: supports listener topic => class configuration

Then we provide a `bin` script to kick it all off...
* `bin/fastly_nsq`

Running this script, given the example configuration, we will end up with a `Launcher` that has one `heartbeat` and one `Manager`. That `Manager` will have started up one `Listener` for the single configured topic. This `Listener` blocks until it gets a message, can be terminated and killed, restarts if needed, and passes messages along to the configured `Processor`.

Notable new "features":
* before, any thread exception cause the whole process to die and restart... There was not much grace to that process. Now, each `Listener` thread has a management path that allows it to die alone and restart alone if needed
* no longer need to use a rake task... though that should still work.


lib/fastly_nsq/version.rb
```diff
@@ -1,3 +1,3 @@
 module FastlyNsq
-  VERSION = '0.9.4'.freeze
+  VERSION = '1.9.4'.freeze
```
ezkl: You may want to reset the minor and patch to 0 since you bumped major.

lib/fastly_nsq/version.rb
```diff
@@ -1,3 +1,3 @@
 module FastlyNsq
-  VERSION = '1.9.4'.freeze
+  VERSION = '0.10.0'.freeze
```
ezkl: 👍

bin/fastly_nsq
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env ruby
+
+require 'pry'
+require 'fastly_nsq/cli'
+require 'dotenv'
+
+Dotenv.load('.env.test')
```
leklund: leftover from testing?

.env.test
```diff
@@ -0,0 +1,13 @@
...
+BROADCAST_ADDRESS='billing'
+NSQD_TCP_ADDRESS='billing:1910'
+NSQD_HTTP_ADDRESS='billing:1911'
+NSQLOOKUPD_TCP_ADDRESS='billing:1918'
+NSQLOOKUPD_HTTP_ADDRESS='billing:1919'
+
+# NSQ TLS Certficate and Key Files
+NSQ_SSL_KEY='/home/vagrant/fastly_cookbooks/fst-nsq/files/default/certs/dev/nsq-client-key.pem'
+NSQ_SSL_CERTIFICATE='/home/vagrant/fastly_cookbooks/fst-nsq/files/default/certs/dev/nsq-client-cert.pem'
+NSQ_SSL_CA_CERTIFICATE='/home/vagrant/fastly_cookbooks/fst-nsq/files/default/certs/dev/nsq-ca-cert.pem'
```
leklund: Probably don't need the `.env.test` checked in

example_config_class.rb
```diff
@@ -0,0 +1,16 @@
...
+end
+
+FastlyNsq.configure do |config|
+  config.channel = 'william'
+  # config.logger = Logger.new
+  # config.preprocessor = ->(message) { FastlyNsq.logger.info 'PREPROCESSESES' }
+
+  config.listen_to do |topics|
+    topics.add('assign_supplemental_plans', ThingWorker)
+  end
```
leklund: This is a little confusing.  `FastlyNsq.configure` will yield FastlyNsq. Then `listen_to` yields `FastlyNsq::Listener::Config.new` (or the existing config). Even charing `topics` here to `listener_config` would help. Or even explicitly calling `FastlyNsq::Listener.config do |listener_config|` might be better. Especially since this is configuring the topics for the listener but isn't actually doing the listening so in that sense `listen_to` is not the best name.

lib/fastly_nsq/cli.rb
```diff
@@ -0,0 +1,238 @@
...
+      FastlyNsq.logger.warn "Thread TID-#{thread.object_id.to_s(36)} #{thread['fastly_nsq_label']}"
+      if thread.backtrace
+        FastlyNsq.logger.warn thread.backtrace.join("\n")
+      else
+        FastlyNsq.logger.warn '<no backtrace available>'
+      end
+    end
+  end
+
+  def damonize
```
leklund: pedantic `daemonize`

lib/fastly_nsq/cli.rb
```diff
@@ -0,0 +1,238 @@
...
+
+  def logfile?
+    !logfile.nil?
+  end
+
+  def pidfile?
+    !pidfile.nil?
+  end
+
+  def damonize?
```
leklund: `daemonize`

lib/fastly_nsq/cli.rb
```diff
@@ -0,0 +1,238 @@
...
+class FastlyNsq::CLI
+  include Singleton
+
+  attr_reader :options
+
+  def parse_options(args = ARGV)
+    parse(args)
+    setup_logger
+    check_pid
+    damonize if damonize?
```
leklund: `daemonize`

lib/fastly_nsq/cli.rb
```diff
@@ -0,0 +1,238 @@
...
+  end
+
+  def logfile
+    options[:logfile]
+  end
+
+  def pidfile
+    options[:pidfile]
+  end
+end
```
leklund, line 238: The cli is really a daemonized listener / multi-threaded consumer. Maybe the name should reflect that?

lib/fastly_nsq/listener.rb
```diff
@@ -1,59 +1,115 @@
...
       @preprocessor = options[:preprocessor]
+      @processor    = processor
+      @thread       = nil
+      @topic        = topic
     end
 
-    def go(run_once: false)
-      exit_on 'INT'
-      exit_on 'TERM'
+    def identity
```
leklund, line 26: This method doesn't seems to be used anywhere?

lib/fastly_nsq/listener.rb
```diff
@@ -1,59 +1,115 @@
...
+    end
+
+    def kill
+      @done = true
+      return unless @thread
+      @logger.info "< Listener KILL: topic #{@topic}"
+      @thread.raise FastlyNsq::Shutdown
+    end
 
-    attr_reader :topic, :consumer, :preprocessor, :processor, :logger
```
leklund: oh rubocop.

lib/fastly_nsq/listener.rb
```diff
@@ -1,59 +1,115 @@
...
+      @done = false
+      @thread = nil
+      self
     end
   end
 end
+
+class FastlyNsq::Shutdown < StandardError; end
+
+require 'fastly_nsq/listener/config'
```
leklund: this seems odd at the bottom of the file? any reason?

lib/fastly_nsq/listener/config.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+module FastlyNsq
+  class Listener::Config
+    attr_reader :listeners
```
leklund: I don't like the name `listeners` here because these aren't really `FastlyNsq::Listener` objects. This is an array of topic->processor class mappings. `processors`? `topics`? `topic_mapping`?

lib/fastly_nsq.rb
```diff
@@ -10,15 +10,51 @@
...
+  end
+
+  def logger=(logger)
     strategy.logger = logger
   end
 
-  def self.logger
+  def self.preprocessor=(preprocessor)
+    @preprocessor ||= preprocessor
+  end
```
leklund, line 25: Why is this here? Aren't preprocessors a pre listener concept? Or is this to setup a global pre-processor?

lib/fastly_nsq/listener.rb
```diff
@@ -1,59 +1,115 @@
...
+        consumer:     @consumer,
+        logger:       @logger,
+        manager:      @manager,
+        preprocessor: @preprocessor,
+        processor:    @processor,
+        topic:        @topic,
+      }
+    end
+
+    def clean_dup
```
leklund: a more descriptive name perhaps? `reset_and_dup`?

bin/fastly_nsq
alieander: oh yeah... oopsy.

.env.test
alieander: a solid point.

example_config_class.rb
jaw6: It seems as if this could be `listen_to topic: Worker` (as in: give a hash, where topics are keys, workers are values). Are there advantages I'm not seeing to the block mechanism here? I also tend to think it's potentially confusing, just curious what the upsides might be.

lib/fastly_nsq.rb
jaw6: Preprocessor hook was intended for common, pre-processor work that applies to all received messages; iirc, the preprocessor hook is mostly used for logging/statsd. If we're moving to daemons, I assume we could get rid of the rake task and preprocessor, since we can expect the upstream client to define *all* relevant processing.

lib/fastly_nsq/listener.rb
alieander: it is used in test...
I know that sounds bad but I used it to test the `cloning bit`...

lib/fastly_nsq/listener.rb
alieander: putting it at the top broke things.
I am sure there are ways to change the load order somewhere else... but the Listener class sees to want to be loaded first...

lib/fastly_nsq/listener/config.rb
alieander: sure.
`topic_map`?

lib/fastly_nsq/listener.rb
alieander: `dupdupdup`
:P 

will change.

example_config_class.rb
alieander: This is pulled from Sidekiq.

The configure part seem to be build up this way so you can have several calls out to configure while init is happening.
```ruby
  config.channel = 'william'
  config.logger = SomeLogger
end

# Some time later
  config.listen_to do |topics|
    topics.add('thing', ThingProcessor)
  end
end
```

The block mechanism for adding to the `topic => Worker` map is again seemingly built up to allow you to additively construct your hash. This hash does not need to stay a hash either... it could be something more useful down the line. You could do this by passing hashes as well and just merging hashes... this is once again a style pulled from PerodicJobs in Sidekiq... You could say that it is slightly easier to validate, log, or do other actions on each added item rather than to loop over a merged hash each time...

alieander: I was going to say that there is the chance you could want to listen to the same topic several times
```ruby
[
  { topic: 'thing', klass: Doer },
  { topic: 'thing', klass: OtherDoer },
]
```
I feel like this will be something we eventually want this to be able to accomplish... However, to have that actually work several messages, you would need different channels. 
```ruby
[
  { topic: 'thing', klass: Doer, channel: 'one' },
  { topic: 'thing', klass: OtherDoer, channel: 'two' },
]
```
That would mean more work here as you would have to push the channel configuration down to the Listener rather than being global. Additionally, you can accommodate that now by just running two daemons...

lib/fastly_nsq/cli.rb
alieander: hmmm...
The intention of this file is to only be used via some `command`... I could see in the future pulling this apart and making the a couple of things? the parsing cli and file checking and then the daemonizing bits as two things...

should I do this now?

bin/fastly_nsq
```diff
@@ -2,6 +2,9 @@
 
 require 'pry'
 require 'fastly_nsq/cli'
+require 'dotenv'
+
+Dotenv.load('.env.local')
```
jaw6: This appears to reverse course on 81d59b7fb

alieander:
Ah shoot

example_config_class.rb
```diff
@@ -7,10 +7,10 @@ def self.process(message)
...
   config.channel = 'william'
-  # config.logger = Logger.new
-  # config.preprocessor = ->(message) { FastlyNsq.logger.info 'PREPROCESSESES' }
+  config.logger = Logger.new
+  config.preprocessor = ->(_) { FastlyNsq.logger.info 'PREPROCESSESES' }
 
-  config.listen_to do |topics|
-    topics.add('assign_supplemental_plans', ThingWorker)
+  config.listener_config do |lc|
+    lc.adld_topic('assign_supplemental_plans', ThingWorker)
```
jaw6: I know this is an example file, but: `add_topic`

example_config_class.rb
jaw6: I like the idea that one could configure multiple processors for a single topic this way, but if it also requires an additional channel, that seems to undercut the utility somewhat.

It occurs to me that there are two configure blocks -- `configure do |config|` and also `listen_to do |topics|` -- that could possibly be made more elegant with just one -- possibly `config.listen_to topic, Processor`.

lib/fastly_nsq/listener.rb
jaw6: If you change `class Listener::Config` to `class Listener\nclassConfig` it should work as expected. (iirc, `class Foo::Bar` assumes/requires that Foo has already been defined) Alternatively, you could move the `Listener::Config` inline in this file: both classes are reasonably small to make that manageable, imo.

alieander: Config was in line initially...

example_config_class.rb
alieander: having two channels is only required if you want to get the same message twice and processes them in different ways... That is an NSQ thing... you could set up two listeners with the same topic and channel and different processors but only the first one to pick up the message would get worked... or both might get the message if timing was correct... but it would not be guaranteed

PR #50, opened 2017-05-22
Consumer should connect on initialize

leklund:
* fixes bug in FastlyNsq::Producer when calling `logger`
* ensures that the consumer's connection to `nsq` is created on initialization.

Recommended Reviewers
=====================
@fastly/billing


alieander:
:+1::dango:

jaw6:
I wouldn't be surprised if we eventually wanted timeout details to be ENV-var configurable, but this seems like a good place to start for now.

PR #51, opened 2017-06-09
Cleanup / Warning removal

alieander:
There where several ruby warnings throw when running the tests.
There are still a few more fro dependancies


leklund:
:+1:

leklund (approved):

PR #52, opened 2017-06-26
Add topic to listener log lines.

leklund:
Reason for Change
===================
* The will make parsing logs much easier 

In the future I wonder if we want to change this log line to `.debug` or make it configurable. When pushing a large volume of messages the logs get rather large. And in most cases I wonder if it wouldn't be better to configure this in the listener rake tasks inside the message processor as part of the application so that we only log the parts we want rather than the entire message.

List of Changes
===============
* add [topic] to the message received log lines in the listener.

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -85,7 +85,7 @@ module FastlyNsq
     private
 
     def log(message)
-      @logger.info "[NSQ] Message Received: #{message}" if @logger
+      @logger.info "[NSQ] Message received on topic [#{@topic}]: #{message}" if @logger
     end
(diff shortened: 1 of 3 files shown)


jaw6:
LGTM :+1:

jaw6:
@leklund and @alieander is this still something we need? any issues merging this now? (I mean, it looks up-to-date, per GH, any other reason to be wary?)

leklund:
Yeah, still would like to have this to makes it easier to filter logs. I think it got forgotten about since we weren't ready to deploy `0.10.0` (#49) yet.

PR #53, opened 2017-08-16
Add accessor for Nsq::Message object and delegate methods

leklund:
Reason for Change
===================
* We'd like to be able to requeue jobs in NSQ vs just dumping them into Sidekiq

List of Changes
===============
* Change the initialization of `FastlyNsq::Message` to accept the `Nsq::Message` instead of just the body.
* add `Forwardable` and delegate methods to Nsq::Message. https://github.com/wistia/nsq-ruby/blob/50f943cf7b58f9d2b4c071accaf46658597d89a2/lib/nsq/frames/message.rb

Risks
=====
* Low. The initialization changes but the public interface remains the same.

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -94,8 +94,9 @@ module FastlyNsq
     end
 
     def next_message
-      message = @consumer.pop # TODO: consumer.pop do |message|
-      result  = yield FastlyNsq::Message.new(message)
+      nsq_message = @consumer.pop # TODO: consumer.pop do |message|
+      message = FastlyNsq::Message.new(nsq_message)
+      result  = yield message
       message.finish if result
     end
 
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -95,7 +95,7 @@ module FastlyNsq
 
     def next_message
       message = @consumer.pop # TODO: consumer.pop do |message|
-      result  = yield FastlyNsq::Message.new(message.body)
+      result  = yield FastlyNsq::Message.new(message)
       message.finish if result
     end
 
--- a/lib/fastly_nsq/message.rb
+++ b/lib/fastly_nsq/message.rb
@@ -3,9 +3,9 @@ require 'json'
 class FastlyNsq::Message
   extend Forwardable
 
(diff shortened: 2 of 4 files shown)


alieander:
Can you give me an example of when you would re-enqueue a job?
Is it the same job as one that is already on the queue?

alieander:
Ah I see. you want to be able to directly access the `.finish`, `.requeue`, etc. methods rather than just having access to the message's body.

alieander (approved):

leklund:
> Ah I see. you want to be able to directly access the .finish, .requeue, etc. methods rather than just having access to the message's body.

Exactly.

Currently the listener and the processors can access only the body of the message. Adding support for the Nsq::Message methods means that a processor has the ability to `finish` or `requeue` an NSQ message. This will, in turn, make it easier to decouple a FastlyNsq::Listener from Sidekiq.

alieander:
They are not really coupled right now. I think that is why I was a bit confused.
The move to calling `.process` over `.perform` made it so you could have any class that implements a `.process` method work as a message "Processor".

set5think:
> They are not really coupled right now. I think that is why I was a bit confused.
The move to calling .process over .perform made it so you could have any class that implements a .process method work as a message "Processor".

The one thing that doesn't make the distinction fully clear (other than language) is that the `.process` is always presumed to be a class level method, rather than `.perform` always presumed to be an instance level method.

leklund:
Per our convo on Slack, I'm going to add an attribute to FastlyNsq::Message that can determine wether the message has been managed, i.e. requeued or finished. And if it has been managed, then the finish and requeue methods are just a no-op.

alieander:
:+1::dango:

leklund:
@alieander so it didn't work with `super` because it is a delegated method but I implemented the `managed` flag here: adc484a

alieander:
I was thinking `@managed` would be a `:symbol` of what had already happened... but that would not be useful yet and it can easily be changed later if needed.

on calling `super`:
Yeah that makes sense. Not a subclass but we both understood what we were getting at :P

leklund:
@alieander made some changes that I think help this make more sense and actually do what we were talking about.

alieander:
Nice! I like that.

PR #54, opened 2017-08-17
This should load rails correctly.

alieander:


leklund (approved):

set5think:
+1

lib/fastly_nsq/cli.rb
```diff
@@ -55,11 +55,20 @@ def read_loop
...
     FastlyNsq.logger.info "Running in #{RUBY_DESCRIPTION}"
     FastlyNsq.logger.info 'Starting processing, hit Ctrl-C to stop' unless options[:daemon]
   end
 
+  def boot_rails
+    return unless ENV['RAILS_ENV']
+
+    require 'rails'
+    require File.expand_path('./config/application.rb')
+    require File.expand_path('./config/environment.rb')
```
alieander, line 69: Hmmm... This is assuming right now that the cli is run out of the root of your rails app. Safe enough assumption but could be configurable...

leklund: Good point, this will obviously break for a sinatra app.

PR #55, opened 2017-09-06
Cleanup in the terminate and kill methods

alieander:
This means that a listener will not ensure it has its connections
terminated anymore but the manager will still do so when appropriate.

Before we where killing terminating the consumer whenever any error occured
and never re-creating it...

This has potential to make the consumer connection live on in some bad cases.
The alternative would be to re-create the Consumer whenever we need to
dup_and_reset.

### DIFF
--- a/spec/lib/fastly_nsq/listener_spec.rb
+++ b/spec/lib/fastly_nsq/listener_spec.rb
@@ -168,10 +168,13 @@ RSpec.describe FastlyNsq::Listener do
         listener.start
         state = listener.instance_variable_get(:@done)
         expect(state).to eq false
+        consumer = listener.instance_variable_get(:@consumer)
+        allow(consumer).to receive(:terminate)
         listener.terminate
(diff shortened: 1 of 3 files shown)


spec/lib/fastly_nsq/listener_spec.rb
```diff
@@ -171,7 +171,7 @@ def self.clear
         listener.terminate
 
         state = listener.instance_variable_get(:@done)
-        expect(logger).to have_received(:info).twice
+        expect(logger).to have_received(:info).thrice
```
leklund, line 176: Can we add an expectation

```
expect(consumer).to have_recived(:terminate)
```

or set the exception on listener to have received `cleanup`?

alieander: good call.
made the tests work but did not make the tests test the new functionality... shame :bell: on me

PR #57, opened 2017-09-15
Only raise if consumer is empty

alieander:
Terminate should allow a consumer to attempt to finish the work that it is
doing. We want it to finish the message it is on and then on the next pass
of the loop, see that @done is set and finish up. However, if there are no
messages, consumer.pop will be blocking and will never find out that @done
is set. We will hit the stop deadline and end up calling kill. Here we are
adding a check to see if the consumer is empty, and only if it is do we
raise to get it out from being stuck on a pop. Otherwise, we allow it to
finish the work it is on and findout that
@done is set ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ block. An exmaple of this can be found here: [`Example Config`](exmaple_config_c
 
 An example of using the cli:
 ```bash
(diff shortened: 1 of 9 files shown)


leklund (approved):
:+1:

PR #58, opened 2017-11-02
remove connection timeout on initialization

alieander:
This is needed for the Producer but not the Consumer.

### DIFF
--- a/lib/fastly_nsq/consumer.rb
+++ b/lib/fastly_nsq/consumer.rb
@@ -13,17 +13,11 @@ module FastlyNsq
       @channel     = channel
       @tls_options = TlsOptions.as_hash(tls_options)
       @connector   = connector
-      Timeout.timeout(5) do
-        sleep(0.1) until connection.connected?
-      end
-    rescue Timeout::Error => error
-      FastlyNsq.logger.error "Consumer for #{topic} failed to connect!"
-      connection.terminate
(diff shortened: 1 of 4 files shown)


lib/fastly_nsq/consumer.rb
```diff
@@ -13,17 +13,11 @@ def initialize(topic:, channel:, tls_options: nil, connector: nil)
...
       @tls_options = TlsOptions.as_hash(tls_options)
       @connector   = connector
-      Timeout.timeout(5) do
-        sleep(0.1) until connection.connected?
-      end
-    rescue Timeout::Error => error
-      FastlyNsq.logger.error "Consumer for #{topic} failed to connect!"
-      connection.terminate
-      raise error
+      connection
```
set5think, line 16: so, now we'll have connections for all topics we've declared we want to listen to, but won't care about their state?

lanej (approved):

lib/fastly_nsq/consumer.rb
alieander: In the consumer it shouldn't matter. You just want to block on `pop` until something comes in. The `Timeout` was not really helping here.

(additionally you don't really return the `connection` here... we are just calling connection to get it to start the `discovery_thread` loop)

set5think (approved):

PR #59, opened 2017-11-27
Implement Most Nsqd and Nsqlookupd Http api's

alieander:
With this there will be a way for any project using this gem to access the http endpoints without a lot of hassle. This means ready access to `stats` in both the text output and a consumable `json` format.

Reason for Change
===================
We want to have better insight into the working of our NSQ nodes. Several useful API's are gated around HTTP access... This provides access to those.

List of Changes
===============
* Add HTTP client `FastlyNsq::Http`
* Add `Nsqd` and `Nsqlookupd` api implementations
* document these endpoints and test them... somwhat

Risks
=====
* The SSL interaction defaults to VERIFY_NONE which is... not great. This will however work in more cases
* This will provide easier access to deletion of whole `topics` and `channels`...

TODO
====
* The `Nsqd` api implements some debuging api's that might be useful in the future... those are not implemented here.
* The binary form of `mpub` needs to be implemented


lib/fastly_nsq/http.rb
```diff
@@ -0,0 +1,57 @@
...
+    req = Net::HTTP::Post.new(uri.request_uri)
+    http.request(req, body)
+  end
+
+  def use_ssl
+    return unless can_use_ssl?
+    http.use_ssl = true
+    http.cert = cert
+    http.key = key
+    http.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
lanej, line 35: should we not verify?

spec/lib/fastly_nsq/http_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      FastlyNsq::Http.new(uri: url).post(data, body)
+
+      expect(a_request(:post, url).with(body: body)).to have_been_requested
+    end
+  end
+
+  describe 'SSL' do
+    it 'can be asked to use SSL' do
+      ssl_url = 'https://example.com:80/boop'
```
lanej, line 54: boop

lib/fastly_nsq/http/nsqlookupd.rb
```diff
@@ -0,0 +1,99 @@
...
+      new(request_uri: '/delete_topic', **args).get(topic: topic)
+    end
+
+    ##
+    # Deletes an existing channel of an existing topic
+    #
+    # @param topic [String] an exsiting topic
+    # @param channel [String] the exsiting channel to delete
+    def self.delete_channel(topic:, channel:, **args)
+      new(request_uri: '/delete_channel', **args).get(topic: topic, channel: channel)
```
lanej: This appears to not match the documentation: http://nsq.io/components/nsqd.html#post-channeldelete.

lanej (approved):
Looks fine to me.  I'm curious about the differences between the [nsqd api](http://nsq.io/components/nsqd.html) and the [nsqlookupd api](http://nsq.io/components/nsqlookupd.html).

lib/fastly_nsq/http.rb
leklund: I thought it was a global sign cert but it might be self signed? Though that's for our specific use case. Maybe this could be configurable?

lib/fastly_nsq/http.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  attr_reader :cert_filename, :key_filename, :uri
+
+  def http
+    @http ||= Net::HTTP.new(uri.host, uri.port)
+  end
+
+  def can_use_ssl?
+    !(cert_filename.nil? || key_filename.nil?)
+  end
```
leklund: we can only use ssl if the cert filename is nil and the key filename is nil? Is that correct?

leklund, line 48: Ignore my now deleted comment. I completely misread the logic here.

leklund (approved):

lib/fastly_nsq/http.rb
alieander: For anything else you would have to make sure the local cert store is setup correct on each box this would be used on. The most annoying case would be for developers however This could easily be made to be configurable.

spec/lib/fastly_nsq/http_spec.rb
alieander: sloop

lib/fastly_nsq/http.rb
alieander: 🍡

lib/fastly_nsq/http.rb
lanej: +1 to configuration

PR #61, opened 2017-11-30
To rubocop update

alieander:
Reason for Change
===================
There is a vulnerability in the version on rubocop we are using I guess.

List of Changes
===============
* update the version of rubocop to the latest release version (0.51.0)
* fix all new things that rubocop now finds offensive
* ignore some of the things rubocop finds offensive
* remove a bunch of cop configurations that did not actually affect the repo
* cleanup
* Add `Josh Lane` to the authors

Risks
=====
No Risks.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,9 +26,6 @@ GuardClause:
 IfUnlessModifier:
   Enabled: false
 
-ClassLength:
-  Enabled: false
-
 Style/ClassVars:
   Description: 'Avoid the use of class variables.'
   StyleGuide: 'https://github.com/bbatsov/ruby-style-guide#no-class-vars'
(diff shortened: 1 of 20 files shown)


.rubocop.yml
```diff
@@ -4,69 +4,55 @@ AllCops:
   Exclude:
     - 'db/schema.rb'
     - 'vendor/**/*'
-  TargetRubyVersion: 2.1
+    - 'Gemfile'
+  TargetRubyVersion: 2.2
```
leklund, line 8: 2.2? Any reason not to target 2.4?

leklund (approved):

.rubocop.yml
alieander: Other than 2.4 increasing the rubocop errors 2 fold...
The Travis builds and tests 2.2, 2.3, and 2.4 so technically that seems to be our oldest supported version

PR #62, opened 2017-12-01
Set metadata

set5think:
Reason for Change
===================
* Allow specifying custom metadata on a message.  We have a meta section on the message, but weren't allowing an accessible way to put data there.

List of Changes
===============
* Allows specifying custom metadata on a message.
* Update changelog and bump version

Risks
=====
* This is a backwards-compatible change.  The two possible risks here are:
  - you can override originating_service if you put that as part of the meta hash you pass
  - if you don't pass in a hash for meta, you'll get a runtime error.

These risks are negligible and will easily be discovered immediately as an operator error

Checklist
=========

- [X] I have NOT linked to any Jira tickets.

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/fastly_nsq.gemspec
+++ b/fastly_nsq.gemspec
@@ -8,7 +8,7 @@ Gem::Specification.new do |gem|
   gem.summary       = 'Fastly NSQ Adapter'
   gem.description   = "Helper classes for Fastly's NSQ Services"
   gem.license       = 'MIT'
-  gem.authors       = ["Tommy O'Neil", 'Adarsh Pandit', 'Joshua Wehner', 'Lukas Eklund', 'Josh Lane']
+  gem.authors       = ["Tommy O'Neil", 'Adarsh Pandit', 'Joshua Wehner', 'Lukas Eklund', 'Josh Lane', 'Hassan Shahid']
   gem.email         = 'tommy@fastly.com'
   gem.homepage      = 'https://github.com/fastly/fastly_nsq'
 
--- a/lib/fastly_nsq/messenger.rb
+++ b/lib/fastly_nsq/messenger.rb
@@ -4,12 +4,12 @@ module FastlyNsq::Messenger
 
   module_function
 
-  def deliver(message:, on_topic:, originating_service: nil)
+  def deliver(message:, on_topic:, originating_service: nil, meta: {})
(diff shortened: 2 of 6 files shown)


lib/fastly_nsq/messenger.rb
```diff
@@ -4,12 +4,12 @@ module FastlyNsq::Messenger
...
   module_function
 
-  def deliver(message:, on_topic:, originating_service: nil)
+  def deliver(message:, on_topic:, originating_service: nil, meta: {})
     payload = {
       data: message,
       meta: {
         originating_service: originating_service || self.originating_service,
-      },
+      }.merge(meta),
```
alieander: don't you want this to merge into the existing `meta` key?

alieander: oh man. that diff confused me.

set5think: hah yea that does look weird in the diff, but it is doing the "right" thing.

leklund: could do
```
meta[:originating_service] = originating_service || self.originating_service
payload =  {
  data: message,
  meta: meta
}
```
and that would prevent overwriting originating service

alieander: or you could raise `NotAllowedToUpdateThatError` :P

fastly_nsq.gemspec
```diff
@@ -8,7 +8,7 @@ Gem::Specification.new do |gem|
   gem.summary       = 'Fastly NSQ Adapter'
   gem.description   = "Helper classes for Fastly's NSQ Services"
   gem.license       = 'MIT'
-  gem.authors       = ["Tommy O'Neil", 'Adarsh Pandit', 'Joshua Wehner', 'Lukas Eklund', 'Josh Lane']
+  gem.authors       = ["Tommy O'Neil", 'Adarsh Pandit', 'Joshua Wehner', 'Lukas Eklund', 'Josh Lane', 'Hassan Shahid']
```
alieander, line 11: There is also a spot for you in the bottom of the readme.

set5think: added

alieander (approved):

PR #63, opened 2017-12-01
Don't log the logger object when creating a listener.

leklund:
Reason for Change
===================
* Because of an interesting interaction between the fluentd logger and is use of thread ID to log `last_error` logs were growing exponentially as new listener threads were created after errors were raised.

List of Changes
===============
Remove the `@logger` object from the `listener.identity` hash.

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -27,7 +27,6 @@ module FastlyNsq
     def identity
       {
         consumer:     @consumer,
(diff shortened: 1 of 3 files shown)


alieander (approved):
This is smart.
Identity is only used for tests and logging so it should not cause any issues.

I do wonder if we should re init the logger when we `reset_then_dup` as otherwise we will end up with the same object...

PR #65, opened 2018-01-03
Consumer read loops send work to a central, prioritized thread pool

lanej:
Reason for Change
===================
* Concurrency can only be controlled by creating multiple bin/fastly_nsq processes.  Now concurrency can be specified and controlled directly by a command line switch (`-c`).
* Listeners now share a common thread pool, greatly improving utilization and performance
* Using `concurrent-ruby` decreases complexity of thread management

List of Changes
===============
* NSQ interface classes respond to `#terminate` and `#connected?`
* A single manager instance is used to track listeners
* Listeners feed work directly from `Nsq::Consumer#read_loop` directly into the manager's thread pool for processing
* Remove `FakeBackend` and `Strategy` references in favor of a `sidekiq` style testing interface e.g. `FastlyNsq::Testing.fake!`
* Remove Rake task support
* Test against a live NSQ cluster
* Add listeners during configuration by using `listen`
* Removes support for delayed processing when testing.

```ruby
  config.listen 'posts', PostListener
  config.listen 'blogs', ->(m) { puts "Blog #{m.body['name']} created!"  }
end
```

Risks
=====
* Performance measurements are pending but the priority_queue_cxx implementation is optimized for `#pop` and `#push`.  We are not modifying an elements priority.  See https://github.com/boborbt/priority_queue_cxx#which-is-the-best-priority-queue-implementation-for-ruby
* Interface changes consitute a major release.

Checklist
=========

Recommended Reviewers
=====================
@fastly/billing


lib/fastly_nsq/consumer.rb
```diff
@@ -1,43 +1,39 @@
...
+    @connect_timeout = connect_timeout
 
-    def connection
-      @connection ||= connector.new(params)
-    end
+    @connection = connect(queue)
+  end
+
+  def empty?
+    connection.size.zero?
```
set5think: pedantic: `size` is delegated to `connection` above.

lib/fastly_nsq/feeder.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def initialize(processor, priority)
+    @processor = processor
+    @priority = priority
+  end
+
+  # @see http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/ThreadPoolExecutor.html#post-instance_method
+  # @see {Nsq::Connection#read_loop}
+  def push(message)
+    FastlyNsq.manager.pool.post(priority) { processor.call(message) }
```
set5think, line 14: processor here would be the class consuming the messages?  If so, any reason to not just leave the existing convention of a method called `process` per processor?

lanej: `processor` here is the instance of `FastlyNsq::Listener`.  I made an 'executive' decision to switch to `#call` so everything is compatible with blocks.  Completely open for discussion but I like the flexibility.

set5think: I agree about the use of `call` moving forward, but at the least I think we should support an alias to process while we're rolling this out.

lanej: I was comfortable changing this now because of the major version bump that is required due to the other changes.  If you feel strongly about it, an alias will definitely work.

lib/fastly_nsq.rb
```diff
@@ -1,62 +1,56 @@
...
+    attr_writer :logger
 
-  def channel=(channel)
-    @channel ||= channel
-  end
-
-  def logger=(logger)
-    strategy.logger = logger
-  end
+    def listen(topic, processor)
```
alieander: should this not also take a `priority` at the very least?

lanej: oooo good call.

lib/fastly_nsq/feeder.rb
alieander: We are really the only users currently of this... with a major bump we are going to have to run around and change a bunch of stuff wherever this is used already. Mine as well make the change now so we can do it all at once!

lib/fastly_nsq.rb
lanej: https://github.com/fastly/fastly_nsq/pull/65/commits/aa3c1e814679373e62a5d2d42a81b76633fb5177

lib/fastly_nsq/listener.rb
```diff
@@ -1,117 +1,53 @@
...
-        processor:    @processor,
-        topic:        @topic,
-      }
-    end
+    @channel      = channel
+    @logger       = logger
+    @preprocessor = preprocessor
+    @processor    = processor
+    @topic        = topic
+    @priority     = priority
```
alieander, line 24: if you had this as `@priority = priority || DEFAULT_PRIORITY` rather than having it in the arguments it could be passed `nil` and still get the default, which could be useful.

lib/fastly_nsq/feeder.rb
set5think: good points; sold.  I do like that we don't have to change the classes and can just change the configuration to have it call .process in a proc.

lib/fastly_nsq/manager.rb
```diff
@@ -1,104 +1,80 @@
...
   end
 
-  def listener_stopped(listener)
-    @plock.synchronize do
-      @listeners.delete listener
+  def add_listener(listener)
+    logger.info { "Listening to topic:'#{listener.topic}' on channel: '#{listener.channel}'" }
+
+    if topic_listeners[listener.topic]
+      logger.warn("topic: #{listener.topic} was added more than once")
```
alieander: this message should be in a block... so it doesn't make the string if we not setup to log at `warning` level.

lib/fastly_nsq/listener.rb
lanej: If `nil` was passed in, it would fail the `Fixnum` validation above.

lib/fastly_nsq/manager.rb
lanej: Good call.  Needs a test oo.

lib/fastly_nsq/messenger.rb
```diff
@@ -1,20 +1,20 @@
...
+
 module FastlyNsq::Messenger
-  DEFAULT_ORIGIN = 'Unknown'.freeze
+  DEFAULT_ORIGIN = 'Unknown'
   @originating_service = DEFAULT_ORIGIN
 
   module_function
 
-  def deliver(message:, on_topic:, originating_service: nil, meta: {})
+  def deliver(message:, topic:, originating_service: nil, meta: {})
```
alieander, line 9: does this need to change with this PR? seems somewhat unrelated...

lanej: It does not.  I took the opportunity to smooth out the inconsistencies in the named parameters, so this matches `FastlyNsq::Listener`, `FastlyNsq::Consumer` and `FastlyNsq::Producer`.

lanej: https://github.com/fastly/fastly_nsq/pull/65/commits/fd4e1d784cda1eb92b7353f203fd062e09d7cf1a

lib/fastly_nsq/listener.rb
```diff
@@ -1,117 +1,53 @@
...
+    preprocessor&.call(message)
+    result = processor.call(message)
+    nsq_message.finish if result
+  end
 
-    def cleanup
-      @consumer.terminate
-      @logger.info "< Consumer terminated: topic [#{@topic}]"
-    end
+  def connected?
```
set5think: pedantic: delegate to consumer.

lib/fastly_nsq/listener.rb
```diff
@@ -1,117 +1,53 @@
...
 
-    def log(message)
-      @logger.info "[NSQ] Message received on topic [#{@topic}]: #{message}" if @logger
-    end
+  def call(nsq_message)
+    message = FastlyNsq::Message.new(nsq_message)
+    log message
+    preprocessor&.call(message)
+    result = processor.call(message)
+    nsq_message.finish if result
```
set5think: I don't think there's a functional difference here, but this is calling `finish` on the nsq_message instance instead of previous being called on the fastlynsq::message instance.

lib/fastly_nsq/manager.rb
```diff
@@ -1,104 +1,80 @@
...
-    return if @done
-    @done = true
+  def topics
+    topic_listeners.keys
+  end
 
-    FastlyNsq.logger.info { 'Terminating quiet listeners' }
-    @listeners.each(&:terminate)
+  def listeners
+    topic_listeners.values
```
set5think: not a big deal, might be worth uniq'ing this or converting to a set.

lanej: That's a good point.  Set might be easiest.

lib/fastly_nsq/manager.rb
```diff
@@ -1,104 +1,80 @@
...
-  def initialize(options = {})
-    @options = options
-    @done = false
-    @listeners = Set.new
-    @plock = Mutex.new
+  def initialize(logger: FastlyNsq.logger, **pool_options)
+    @done      = false
+    @logger    = logger
+    @pool      = FastlyNsq::PriorityThreadPool.new(
+      { fallback_policy: :caller_runs, max_threads: DEFAULT_POOL_SIZE }.merge(pool_options),
```
set5think, line 13: are we comfortable with `fallback_policy` being overridden by pool_options?

lib/fastly_nsq/consumer.rb
set5think: meaning, can't this be `size.zero?`

lib/fastly_nsq/manager.rb
lanej: You could do some fancy things with the thread pool so I wanted to leave the option open for other consumers.  Definitely advanced territory though.

lib/fastly_nsq/manager.rb
```diff
@@ -1,104 +1,80 @@
...
+  def stop_processing(deadline)
+    logger.info { 'Stopping processors' }
+    pool.shutdown
 
-    FastlyNsq.topic_map.each_pair do |topic, processor|
-      @listeners << setup_listener(topic, processor)
-    end
-  end
+    logger.info { 'Waiting for processors to finish...' }
+    return if pool.wait_for_termination(deadline)
```
set5think, line 69: does `wait_for_termination` return a boolean?  If it returns true, what does that mean for the `transfer` method above?  it seems like transfer will continue assuming `stop_processing` is done.

lanej: Yeah [wait_for_termination](http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/ThreadPoolExecutor.html#wait_for_termination-instance_method) returns `true` if the threads were terminated within the timeout, `false` otherwise.

lib/fastly_nsq/messenger.rb
alieander: This is another thing that is going to need to be updated in a lot of places.
eg: Northstar doesn't have a listener ATM but it does use `Messenger` so, with this change we will also likely want to update Northstar right away just for this (so we don't forget about it later :P )

README.md
```diff
@@ -107,84 +126,51 @@ topic     = 'user_created'
...
-
 ### `FastlyNsq::CLI`
 
 To help facilitate running the `FastlyNsq::Listener` in a blocking fashion
 outside your application, a `CLI` and bin script [`fastly_nsq`](bin/fastly_nsq)
 are provided.
 
 This can be setup ahead of time by calling `FastlyNsq.configure` and passing
-block. An exmaple of this can be found here: [`Example Config`](exmaple_config_class.rb)
+block. An exmaple of this can be found here: [`Example Config`](example_config_class.rb)
```
alieander: lol

example_config_class.rb
```diff
@@ -13,8 +13,6 @@ def self.process(message)
   config.logger = Logger.new
   config.preprocessor = ->(_) { FastlyNsq.logger.info 'PREPROCESSESES' }
 
-  config.listener_config do |lc|
-    lc.add_topic('some_topic', MessageProcessor)
-    lc.add_topic('some_other_topic', MessageProcessor)
-  end
+  lc.listen('some_topic', MessageProcessor)
+  lc.listen('some_other_topic', MessageProcessor, priority: 7)
```
alieander: It would be nice to suggest the usage of a lambda here now that it is possible to do that. Mostly because it is nice and this is OO :P
```
lc.listen('some_puts_topic', -> (msg) { puts msg }, priority: 7)
```

spec/support/env_helpers.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module EnvHelpers
   def use_fake_connection
     allow(ENV).to receive(:[]).with('FAKE_QUEUE').and_return(true)
```
alieander: if you removed the need for `FAKE_QUEUE` is this helper still needed?

spec/producer_spec.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe FastlyNsq::Producer do
+  let!(:topic) { 'fnsq' }
```
alieander, line 6: i did know `let!` is a thing and I love it.
  Is there something that resets or changes `topic` that it needs to be force reset before each test?

spec/producer_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  subject { FastlyNsq::Producer.new(topic: topic) }
+
+  before { reset_topic(topic) }
```
alieander, line 9: force reset here and in the `let!`?

lib/fastly_nsq/manager.rb
set5think: ah, makes sense now. terminate gracefully or force terminate.

lib/fastly_nsq/producer.rb
```diff
@@ -1,41 +1,52 @@
...
-    end
+      ),
+    )
+
+    Timeout.timeout(connect_timeout) { sleep(0.1) until connection.connected? }
+
+    true
+  rescue Timeout::Error => error
+    FastlyNsq.logger.error "Producer for #{topic} failed to connect!"
+    connection.terminate
```
set5think: probably better to call `terminate` here, as defined above?

spec/support/env_helpers.rb
lanej: Nope.  I will dump this file.

example_config_class.rb
lanej: I'll add it to the README

spec/producer_spec.rb
lanej: I'm not seeing the reset inside a let!, can you link it?

lib/fastly_nsq/testing.rb
```diff
@@ -0,0 +1,239 @@
...
+      def reset!
+        return unless enabled?
+        FastlyNsq::Messages.messages.clear
+      end
+    end
+  end
+
+  module Messages
+    def self.messages
+      @messages ||= Hash.new { |h, k| h[k] = [] }
```
set5think, line 59: pedantic: `Hash.new([])`

spec/producer_spec.rb
lanej: The reset is to ensure a clean state in nsq and to fix consistency issues I was having with the test suite.

spec/producer_spec.rb
alieander: Perhaps I misinterpreted this block of text from the rspec tests.
```
Note that let is lazy-evaluated: it is not evaluated until the first time
the method it defines is invoked. You can use let! to force the method's
invocation before each example.
```

nm

lib/fastly_nsq/version.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module FastlyNsq
-  VERSION = '0.13.2'.freeze
+  VERSION = '0.13.2'
```
set5think, line 4: waiting to bump this?

README.md
```diff
@@ -107,84 +126,51 @@ topic     = 'user_created'
...
 
-This will pop the next message
+This will send messages through `FastlyNsq.manager.pool`
 off of the queue
 and send the JSON text body
-to `MessageProcessor.process(message_body, topic)`.
+to `MessageProcessor.call(message)`.
 
-To initiate a blocking loop to process messages continuously:
+Specify a topic priority by providing a number (default is 5)
```
set5think: DEFAULT_PRIORITY is set to 0, unless I'm missing something.

lib/fastly_nsq/version.rb
lanej: I'll bump and tag master after merge.

README.md
lanej: Nope, you are right.  I read `DEFAULT_CONNECTION_TIMEOUT` instead.

lib/fastly_nsq/testing.rb
lanej: I like it.

```
= Hash.new

(from ruby core)
------------------------------------------------------------------------------
  Hash.new                          -> new_hash
  Hash.new(obj)                     -> new_hash
  Hash.new {|hash, key| block }     -> new_hash

------------------------------------------------------------------------------

Returns a new, empty hash. If this hash is subsequently accessed by a key that
doesn't correspond to a hash entry, the value returned depends on the style of
new used to create the hash. In the first form, the access returns nil. If
obj is specified, this single object will be used for all
default values. If a block is specified, it will
be called with the hash object and the key, and should return the default
value. It is the block's responsibility to store the value in the hash if
required.

  h = Hash.new("Go Fish")
  h["a"] = 100
  h["b"] = 200
  h["a"]           #=> 100
  h["c"]           #=> "Go Fish"
  # The following alters the single default object
  h["c"].upcase!   #=> "GO FISH"
  h["d"]           #=> "GO FISH"
  h.keys           #=> ["a", "b"]

  # While this creates a new default object each time
  h = Hash.new { |hash, key| hash[key] = "Go Fish: #{key}" }
  h["c"]           #=> "Go Fish: c"
  h["c"].upcase!   #=> "GO FISH: C"
  h["d"]           #=> "Go Fish: d"
  h.keys           #=> ["c", "d"]
```

lanej: The documentation here is poor but `Hash.new([])` has a different behavior.

>  If obj is specified, this single object will be used for all default values.

```ruby
[1] pry(main)> h = Hash.new([])
=> {}
[2] pry(main)> h['foo'] << 'bar'
=> ["bar"]
[3] pry(main)> h
=> {}
```

> If a block is specified, it will be called with the hash object and the key, and should return the default value. It is the block's responsibility to store the value in the hash if required.

```ruby
[4] pry(main)> h = Hash.new { |h,k| h[k] = [] }
=> {}
[5] pry(main)> h['foo'] << 'bar'
=> ["bar"]
[6] pry(main)> h
=> {"foo"=>["bar"]}
```

set5think: umm, yea that is striking.  Carry on.

set5think:
pending CI, +1 from me, but I think there are still some comments from @alieander.

lib/fastly_nsq/messenger.rb
lanej: Hopefully `find . -iname '*.rb' | xargs sed -i '' 's/on_topic:/topic:'` will work

lanej:
@alieander have I addressed your concerns?

 > tommy says 👍

PR #67, opened 2018-01-05
fix: launcher heartbeat

lanej:
Reason for Change
===================
* Fix broken heartbeat log lines

List of Changes
===============
* Refactor the launcher for testing.  
* Heartbeat messages contains useful information about the pool

```
D, [2018-01-05T11:06:00.448682 #39195] DEBUG -- : HEARTBEAT: busy: 0 processed: 0 max_threads: 3 max_queue_size: 0 listeners: 1
```

Risks
=====
* None

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/launcher.rb
+++ b/lib/fastly_nsq/launcher.rb
@@ -4,23 +4,25 @@ require 'fastly_nsq/safe_thread'
 
 class FastlyNsq::Launcher
   include FastlyNsq::SafeThread
-  extend Forwardable
 
-  attr_reader :timeout
+  attr_reader :timeout, :logger
+  attr_accessor :pulse
 
   def manager
     FastlyNsq.manager
   end
 
-  def initialize(timeout: 5, **options)
+  def initialize(timeout: 5, pulse: 5, logger: FastlyNsq.logger, **options)
     @done    = false
     @timeout = timeout
+    @pulse   = pulse
+    @logger  = logger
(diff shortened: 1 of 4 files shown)


lib/fastly_nsq/launcher.rb
```diff
@@ -1,26 +1,29 @@
 # frozen_string_literal: true
 
+#
```
alieander: ?

lib/fastly_nsq/launcher.rb
```diff
@@ -40,25 +43,28 @@ def stopping?
...
+    logger.debug do
       [
         'HEARTBEAT:',
-        'thread_status:', @manager.listeners.map(&:status).join(', '),
-        'listener_count:', @manager.listeners.count
+        'busy:', manager.pool.length,
+        'processed:', manager.pool.completed_task_count,
+        'max_threads:', manager.pool.max_length,
+        'max_queue_size:', manager.pool.largest_length,
+        'listeners:', manager.listeners.count
```
alieander, line 52: This is great!
I wonder if there is interesting information on the `Priorities` that we can add in in the future.

alieander (approved):

lib/fastly_nsq/launcher.rb
lanej: We definitely could

PR #68, opened 2018-02-16
upgrade nsq-ruby to ~> 2.2

leklund:
Reason for Change
===================
* 2.2 includes a fix for memory leakage (https://github.com/wistia/nsq-ruby/pull/42)

A simple `bundle update nsq-ruby` would work in most cases but this will ensure the version w/o memory leaks is used in the future.

List of Changes
===============
* Update Gemspec

Risks
=====
* None.

Recommended Reviewers
=====================
@fastly/internal-engineering

### DIFF
--- a/fastly_nsq.gemspec
+++ b/fastly_nsq.gemspec
@@ -29,6 +29,6 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency 'webmock', '~> 3.0'
 
   gem.add_dependency 'concurrent-ruby', '~> 1.0'
-  gem.add_dependency 'nsq-ruby', '~> 2.0', '>= 2.0.5'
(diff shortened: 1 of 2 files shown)


set5think (approved):

lanej:
nobody likes memory leaks.  10k buffer limit is interesting though.  We don't actually care about this because we override the queue when we construct listeners.

leklund:
> 10k buffer limit is interesting though. We don't actually care about this because we override the queue when we construct listeners.

Yeah seems like it should be configurable. Maybe I'll do a quick PR to make that an option even though we don't currently need it.

leklund:
> We don't actually care about this because we override the queue when we construct listeners.

This 10k limit looks like it only for the write queue which wouldn't affect the listeners. either way, we're good

PR #70, opened 2018-04-11
global max_attempts setting

lanej:
Reason for Change
===================

`max_attempts`, like `logger` and `channel`, are extra useful with global defaults

List of Changes
===============

All `Consumer` and `Listener` `max_attempts` values will default to `FastlyNsq.max_attempts`

Risks
=====

None.

Checklist
=========

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq.rb
+++ b/lib/fastly_nsq.rb
@@ -14,6 +14,7 @@ module FastlyNsq
   class << self
     attr_accessor :channel
     attr_accessor :preprocessor
+    attr_accessor :max_attempts
     attr_writer :logger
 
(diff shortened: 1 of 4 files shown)


set5think:
Would `nil` be the value for not having any max attempts?

set5think (approved):
One question; but looks good otherwise.

lanej:
@set5think yup

https://github.com/wistia/nsq-ruby/blob/c6ff6e506888067d3bdb3726799264bb9e7dfe4e/lib/nsq/connection.rb#L35

`nil` value appears to be "no max attempts"

PR #71, opened 2018-04-11
standardize logging format and add attempts

lanej:
Reason for Change
===================

* Seeing `attempts` on message received is very useful information
* Standardizing logging format makes log parsing much easier

List of Changes
===============

* Add `attempts` to message received log line
* Use a similar format in log messages

Risks
=====

It's possible consumers with tight log integrations may break

Checklist
=========

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -38,7 +38,19 @@ class FastlyNsq::Listener
 
   def call(nsq_message)
     message = FastlyNsq::Message.new(nsq_message)
-    logger.info "topic #{topic}, attempts #{nsq_message.attempts} received: #{message}"
+
+    logger.info do
+      {
+        channel:  channel,
+        topic:    topic,
+        attempts: nsq_message.attempts,
+        id:       Digest::MD5.hexdigest(nsq_message.body.to_s),
+        metadata: message.meta,
+      }.tap do |l|
+        l[:data] = message.body if logger.level == Logger::DEBUG
+      end
+    end
+
     preprocessor&.call(message)
     result = processor.call(message)
     message.finish if result
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -38,7 +38,7 @@ class FastlyNsq::Listener
 
   def call(nsq_message)
     message = FastlyNsq::Message.new(nsq_message)
-    logger.info "[NSQ] Message received on topic [#{topic}]: #{message}"
+    logger.info "topic #{topic}, attempts #{nsq_message.attempts} received: #{message}"
     preprocessor&.call(message)
     result = processor.call(message)
     message.finish if result
@@ -48,6 +48,6 @@ class FastlyNsq::Listener
   def terminate
     return unless connected?
     consumer.terminate
-    logger.info "< Consumer terminated: topic [#{topic}]"
+    logger.info "topic #{topic}, channel #{channel}: consumer terminated"
   end
 end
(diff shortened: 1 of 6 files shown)


lib/fastly_nsq/listener.rb
```diff
@@ -38,7 +38,7 @@ def initialize(topic:, processor:, preprocessor: FastlyNsq.preprocessor, channel
 
   def call(nsq_message)
     message = FastlyNsq::Message.new(nsq_message)
-    logger.info "[NSQ] Message received on topic [#{topic}]: #{message}"
+    logger.info "topic #{topic}, attempts #{nsq_message.attempts} received: #{message}"
```
alieander: not that this matter that much but we should use the block format in all these `info`'s as well:
ie: `logger.info { "thing" }`

alieander: if we are standardizing,  should this not be:
`"topic #{topic}, channel #{channel}, attempts #{nsq_message.attempts}: received #{message}"`
?

lib/fastly_nsq/manager.rb
```diff
@@ -43,10 +43,10 @@ def stopped?
   end
 
   def add_listener(listener)
-    logger.info { "Listening to topic:'#{listener.topic}' on channel: '#{listener.channel}'" }
+    logger.info { "topic #{listener.topic}, channel #{listener.channel} listening" }
```
alieander: `"topic #{listener.topic}, channel #{listener.channel}: listening"`

only difference is the `:`

alieander:
"topic #{topic}, channel #{channel}, attempts #{nsq_message.attempts}: received #{message}"
seems like the format your aiming for is:
```
topic TOPIC_NAME[, channel CHANNEL_NAME][..., _attr_ VALUE]: ACTION_OR_EVENT[ MESSAGE]
```
with caps names begin variable values and `[ ]` wrapped bits being optional...

Does that sound about right?

lib/fastly_nsq/manager.rb
```diff
@@ -43,10 +43,10 @@ def stopped?
...
 
     if topic_listeners[listener.topic]
-      logger.warn { "topic: #{listener.topic} was added more than once" }
+      logger.warn { "topic #{listener.topic} was added more than once" }
```
alieander: in a matching format could be
`"topic #{listener.topic}: duplicate topic"`

leklund:
If we use the updated generic logger, we can send a hash and then have it parsed in Sumo for us.

```ruby
{
  channel: channel,
  topic: topic,
  message_id: message.nsq_message.id, # possibly usfeul(?) 
  data: severity == debug ? message.data : nil,
  metadata: message.meta,
}
```

lanej:
@alieander yeah you've got the format well defined.  I'll update

@leklund is this too low to assume Timber as a logger?

leklund:
> @leklund is this too low to assume Timber as a logger?

Possibly?  If we are concerned about sending the hash to the logger I suppose we could do something like `FastlyNsq.log_format = :hash` . Though it doesn't seem problematic to send a hash to a basic Logger:

```
[1] pry(main)> Logger.new(STDOUT).warn({ msg: 'foo', meta: { time: 'now', num: 42 } })
W, [2018-04-12T21:59:51.523713 #464375]  WARN -- : {:msg=>"foo", :meta=>{:time=>"now", :num=>42}}
=> true
```

lanej:
updated.

alieander (approved):
Looks good.

PR #72, opened 2018-04-12
add exponential retries to FastlyNsq::Message

leklund:
Reason for Change
===================
* We have several different places where we implement some message retry logic. Rather than adding exponential retries everywhere it makes sense to have it on the FastlyNsq::Message.

Risks
=====
* If max_attempts is not set you could end up with some VERY long requeue periods.

Checklist
=========

Recommended Reviewers
=====================
@fastly/internal-engineering

### DIFF
--- a/lib/fastly_nsq/message.rb
+++ b/lib/fastly_nsq/message.rb
@@ -34,10 +34,17 @@ class FastlyNsq::Message
     nsq_message.finish
   end
 
-  def requeue(*args)
+  def requeue(timeout = nil)
     return managed if managed
+    timeout ||= requeue_period
 
     @managed = :requeued
-    nsq_message.requeue(*args)
+    nsq_message.requeue(timeout)
+  end
+
+  private
+
+  def requeue_period
+    ((attempts**4) + 45 + (rand(60) * (attempts + 1))) * 1_000
   end
 end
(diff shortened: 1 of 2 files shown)


set5think (approved):

lib/fastly_nsq/message.rb
```diff
@@ -34,10 +34,17 @@ def finish
...
 
     @managed = :requeued
-    nsq_message.requeue(*args)
+    nsq_message.requeue(timeout)
+  end
+
+  private
+
+  def requeue_period
+    ((attempts**4) + 45 + (rand(60) * (attempts + 1))) * 1_000
```
alieander, line 48: I feel like this should have a `max` value...

leklund: probably not a bad idea. By attempt 100 the requeue period is > 3 years (attempt 200 is > 50 years)

leklund: capping attempts in the formula to 30 would max out at ~ 9.4 days

lanej: `Nsq::Connection` has a similar formula

https://github.com/wistia/nsq-ruby/blob/c6ff6e506888067d3bdb3726799264bb9e7dfe4e/lib/nsq/connection.rb#L421-L427

They do limit the maximum sleep time.  You have some control over this indirectly by limiting `max_attempts`.

leklund: https://github.com/fastly/fastly_nsq/pull/73

PR #75, opened 2018-04-24
Add `sent_at` to `meta` when `deliver` is called.

set5think:
Reason for Change
===================

With append-only data; timestamps are critical to be accurate and granular.  Because we can't guarantee that every sender of data onto NSQ is going to have all their timestamps be granular before being converted to JSON (which can lose granularity), we ensure a timestamp that at least tells us when this was trying to be delivered with granularity up to the 100,000th.

List of Changes
===============
* Add `sent_at` to `meta` in `deliver`
* Add Timecop for testing.

Risks
=====
* This basically prevents a `sent_at` from being added to the meta section of a message by a sender.

Checklist
=========

- [X] I have NOT linked to any Jira tickets.

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/spec/messenger_spec.rb
+++ b/spec/messenger_spec.rb
@@ -8,12 +8,14 @@ RSpec.describe FastlyNsq::Messenger do
   let(:message)  { { sample: 'sample', message: 'message' } }
   let(:producer) { double 'FastlyNsq::Producer', write: nil, terminate: :terminated }
   let(:origin)   { 'originating_service' }
+  let(:sent_at)  { Time.now.iso8601(5) }
 
   let(:expected_attributes) do
     {
       data: message,
       meta: {
         originating_service: 'originating_service',
+        sent_at: sent_at,
       },
     }
   end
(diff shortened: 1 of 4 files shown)


spec/messenger_spec.rb
```diff
@@ -25,6 +27,8 @@
   end
 
   describe '#deliver' do
+    before { Timecop.freeze(sent_at) }
```
leklund, line 30: Timecop is a perfectly valid dependency to add for testing but you can also do

```
time_now = Time.now
allow(Time).to receive(:now).and_return(time_now)
```

leklund (approved):
one comment re Timecop, but this looks good.

alieander (approved):

spec/messenger_spec.rb
set5think: reasonable; though the one advantage with Timecop is that it will deal with methods that aren't just `now`.

PR #81, opened 2018-05-31
Support writing multiple messages in FastlyNsq::Messenger

leklund:
Reason for Change
===================
`Nsq::Producer` supports writing messages using `mpub` when multiple arguments are passed to `write`.  

This update adds support to multiple messages in `Fastly::Messenger` using `.deliver_multi`

List of Changes
===============
* add new method `Fastly::Messenger.deliver_multi` that takes an array of messages
* update `Nsq::Producer` to splat the arg list that is sent to `write` so that we can support passing in an array.
* Documentation updates.

Risks
=====
* Low. This is a new method and the implementation of `deliver` did not change.

Checklist
=========

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,13 @@ You can also set the originating service for all `deliver` calls:
 FastlyNsq::Messenger.originating_service = 'some awesome service'
 ```
 
+`FastlyNsq::Messenger` also spuports delivering multiple message at once and will
+use the NSQ `mpub` directive under the hood.
+
+```ruby
+FastlyNsq::Messenger.deliver_multi(messages: array_of_msgs, topic: 'my_topic')
+```
+
 `FastlyNsq::Messenger` can also be used to manage Producer connections
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -291,6 +291,19 @@ NSQLOOKUPD_HTTP_ADDRESS='127.0.0.1:4161, 10.1.1.101:4161'
 See the [`.sample.env`](examples/.sample.env) file
 for more detail.
 
+## Development
+
+The fastest way to get up and running for development is to use
+the Docker container provided by Docker Compose:
+
+* Clone: `git clone https://github.com/fastly/fastly_nsq.git`
+* `cd fastly_nsq`
+* run `bundle install`
+* run `docker-compose up -d`
+* `rake spec`
+
+You will still need the `ENV` variables as defined above.
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -291,6 +291,19 @@ NSQLOOKUPD_HTTP_ADDRESS='127.0.0.1:4161, 10.1.1.101:4161'
...
+The fastest way to get up and running for development is to use
+the Docker container provided by Docker Compose:
+
+* Clone: `git clone https://github.com/fastly/fastly_nsq.git`
+* `cd fastly_nsq`
+* run `bundle install`
+* run `docker-compose up -d`
+* `rake spec`
+
+You will still the `ENV` variables as defined above.
```
set5think: s/still/still need/

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       - "4151:4151"
   nsqadmin:
     image: nsqio/nsq
-    command: /nsqadmin --lookupd-http-address=nsqlookupd:4161
+    command: /nsqadmin --lookupd-http-address=nsqlookupd:4171
```
set5think: conflicting with something locally?

lib/fastly_nsq/messenger.rb
```diff
@@ -16,22 +16,52 @@ module FastlyNsq::Messenger
...
+  # @param originating_service [String] added to meta key of message payload
+  # @param meta [Hash]
+  # @return [Void]
+  # @example
+  #   FastlyNsq::Messenger.deliver_multi(
+  #     messages: [{a: 1, count: 11}, {a: 2, count: 22}],
+  #     topic: 'counts',
+  #   )
+  def deliver_multi(messages:, topic:, originating_service: nil, meta: {})
+    meta = populate_meta(originating_service: originating_service, meta: meta)
```
set5think, line 63: the irony in how meta this is...

docker-compose.yml
leklund: oh nm. this is totally unnecessary and would break nsqadmin. I can't even docker.

set5think (approved):

leklund:
We had some discussions on Slack about just supporting arrays in `.deliver`. The only reason for not changing that interface is that in some cases we possible do want to send an array as the payload and don't want to split it into individual messages.

PR #82, opened 2018-06-06
require pry only if pry is available already

alieander:
We do not need this for the CLI at all... its just a nice to have.
We are also not declaring this as a dependancy so we can not just assume
that it will exist.

Reason for Change
===================
requiring pry when it is not around causes crashes...
We are not going to add pry as a dep.

List of Changes
===============
Wrap `require 'pry'` in a being; rescue LoadError; end

Risks
=====
none.

Checklist
=========

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/bin/fastly_nsq
+++ b/bin/fastly_nsq
@@ -1,7 +1,6 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require 'pry'
 require 'fastly_nsq/cli'
 
 begin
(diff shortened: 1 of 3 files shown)


leklund:
we do load `pry` in development because of

```
  gem.add_development_dependency 'pry-byebug', '~> 3.3'
```

The console uses `irb` though because we don't specify that via `Gem::Tasks`

alieander:
We where getting errors when trying to just gem install `fastly_nsq` and then use it... It blows up in the case where your gems are all clean and shit.

leklund:
oh yeah, because that's in the cli.

bin/fastly_nsq
```diff
@@ -1,7 +1,11 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require 'pry'
+begin
+  require 'pry'
+rescue LoadError
+  STDOUT.puts 'Pry not available. Not loading Pry.'
```
leklund: Why do we even care? seems like we can just remove the `require` altogether no?

alieander: we don't and you are correct

PR #85, opened 2018-07-23
add max_req_timeout [IE-2040]

leklund:
Reason for Change
===================

Helps ensure we don't requeue messages
with a timeout larger than the nsqd conifgured max_timeout which will
result in an error like

```
E_INVALID REQ timeout 6836000000000 out of range 0-3600000000000
```

This setting can be configured via the `attr_writer` or via an ENV variable. It defaults to 1 hour.

List of Changes
===============
* add attr_writer for `max_req_timeout` and reader method that memoizes the default.
* update the `FastlyNsq::Message#requeue` to support `max_req_timeout` and only send the a timeout value less than or equal to the max.

Risks
=====
* Low. 

Checklist
=========

Recommended Reviewers
=====================
@fastly/internal-engineering

### DIFF
--- a/spec/message_spec.rb
+++ b/spec/message_spec.rb
@@ -69,4 +69,10 @@ RSpec.describe FastlyNsq::Message do
 
     subject.requeue
   end
+
+  it 'uses the FastlyNsq.max_req_timeout it timeout is larger than FastlyNsq.max_req_timeout' do
+    expect(nsq_message).to receive(:requeue).with(60 * 60 * 1_000)
+
+    subject.requeue(60 * 60 * 4 * 1_000)
+  end
 end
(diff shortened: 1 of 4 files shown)


spec/message_spec.rb
```diff
@@ -69,4 +69,10 @@
 
     subject.requeue
   end
+
+  it 'uses the FastlyNsq.max_req_timeout it timeout is larger than FastlyNsq.max_req_timeout' do
+    expect(nsq_message).to receive(:requeue).with(60 * 60 * 1_000)
+
+    subject.requeue(4.hours * 1_000)
```
lanej: `4.hours` :(.  no active support.

leklund: oops. fixed.

lanej (approved):

PR #90, opened 2018-07-26
Add NewRelic support

leklund:
Reason for Change
===================
* NewRelic transaction tracing and error reporting listeners is desired.

List of Changes
===============
* add Fastly::NewRelic
* automate tracing and error reporting if `NewRelic` is loaded. 

Risks
=====
* Low. NewRelic instrumentation should be invisible.

Checklist
=========

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -113,21 +113,26 @@ class FastlyNsq::Listener
   def call(nsq_message)
     message = FastlyNsq::Message.new(nsq_message)
 
+    msg_info = {
+      channel:  channel,
+      topic:    topic,
+      attempts: nsq_message.attempts,
+      id:       Digest::MD5.hexdigest(nsq_message.body.to_s),
+      metadata: message.meta,
+    }
+
     logger.info do
-      {
-        channel:  channel,
-        topic:    topic,
-        attempts: nsq_message.attempts,
-        id:       Digest::MD5.hexdigest(nsq_message.body.to_s),
-        metadata: message.meta,
-      }.tap do |l|
+      msg_info.tap do |l|
         l[:data] = message.body if logger.level == Logger::DEBUG
       end
     end
 
-    preprocessor&.call(message)
-    result = processor.call(message)
-    message.finish if result
+    FastlyNsq.tracer.trace_with_newrelic(msg_info) do
+      preprocessor&.call(message)
+      result = processor.call(message)
+      message.finish if result
+    end
+
     message
   end
 
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -127,7 +127,7 @@ class FastlyNsq::Listener
       end
     end
 
-    FastlyNsq.tracer.trace_with_newrelic(msg_info) do
+    FastlyNsq.tracer.trace_with_newrelic(params: msg_info, class_name: processor.class.to_s) do
       preprocessor&.call(message)
       result = processor.call(message)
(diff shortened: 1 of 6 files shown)


lib/fastly_nsq/new_relic.rb
```diff
@@ -0,0 +1,31 @@
...
+  include NewRelic::Agent::Instrumentation::ControllerInstrumentation if defined?(::NewRelic)
+
+  attr_reader :agent
+
+  def initialize(agent = NewRelic::Agent)
+    @agent = agent
+  end
+
+  def enabled?
+    @enabled ||= Object.const_defined?('NewRelic')
```
lanej: did `if defined?(::NewRelic)` not work here?

Maybe we should 

```ruby
begin
  require 'new_relic'
rescue LoadError
end
```

to avoid timing issues.

I think this would also simplifying testing.  What do you think?

spec/new_relic.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+module FakeRelic
```
lanej: lol `FakeRelic`.

lib/fastly_nsq/new_relic.rb
leklund: `defined?` doesn't return a boolean which is annoying. That `?` is misleading.

Good call. I'll add the `require; rescue;`

lib/fastly_nsq/listener.rb
```diff
@@ -113,21 +113,26 @@ def initialize(topic:, processor:, preprocessor: FastlyNsq.preprocessor, channel
...
     end
 
-    preprocessor&.call(message)
-    result = processor.call(message)
-    message.finish if result
+    FastlyNsq.tracer.trace_with_newrelic(msg_info) do
+      preprocessor&.call(message)
+      result = processor.call(message)
+      message.finish if result
+    end
```
leklund: Do we want to trace every `processor.call`? I'm wondering about any potential performance issues.

leklund: I know it's in a background thread etc etc, just wondering

set5think: as opposed to random sampling?  I think we do want every processor call traced, since data values/differences can change the characteristics of what happens.

leklund: I was just thinking about massive backfills and if we are adding 5% overhead to every `processor.call` it might be something to think about or be able to toggle.

leklund: @set5think per our discussion I think this is good to add around every call. A next step might be a toggle that allows listeners to determine when to trace. Would be cool to have it be`on/off/sample`

set5think: yea that's a nice idea.

set5think (approved):

lib/fastly_nsq/listener.rb
```diff
@@ -113,21 +113,26 @@ def initialize(topic:, processor:, preprocessor: FastlyNsq.preprocessor, channel
...
-      }.tap do |l|
+      msg_info.tap do |l|
         l[:data] = message.body if logger.level == Logger::DEBUG
       end
     end
 
-    preprocessor&.call(message)
-    result = processor.call(message)
-    message.finish if result
+    FastlyNsq.tracer.trace_with_newrelic(msg_info) do
```
leklund: oh, realized that this will include the body if debug mode and  we don't want that. will update

PR #92, opened 2018-08-13
set Nsq.logger when setting logger

leklund:
Reason for Change
===================
If we don't set `Nsq.logger` it's logs to /dev/null
https://github.com/wistia/nsq-ruby/blob/8774b0a8a06389f3664e86d1d5263807a51aca9d/lib/nsq/logger.rb#L3

Also set the default logger to `Logger.new(STDERR)` 

Risks
=====
* Low. adds logging setup to Nsq.logger so we have visibility there

Checklist
=========

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq.rb
+++ b/lib/fastly_nsq.rb
@@ -30,9 +30,6 @@ module FastlyNsq
     # @return [Integer]
     attr_writer :max_req_timeout
 
-    # @return [Logger]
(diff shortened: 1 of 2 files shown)


alieander (approved):
Looks good.
We should probably default to logging rather than default to not logging.

PR #93, opened 2018-10-02
Consumer still attempts connections in Test mode

alieander:
Reason for Change
===================
* patch out connect so that we are no logger going to pointlessly attempt
to connect to some server when we do not care to expect it to connect.

List of Changes
===============
* add connect method to Testing Consumer

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/testing.rb
+++ b/lib/fastly_nsq/testing.rb
@@ -284,8 +284,7 @@ module FastlyNsq
 
     def connect(*args)
       return super(*args) unless FastlyNsq::Testing.enabled?
-      @connected = true
-      FakeConnection.new
+      @connected = FakeConnection.new
     end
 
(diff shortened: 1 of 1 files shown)


lanej (dismissed):
whoops.

alieander:
@lanej added a commit to this that I am not proud of... but I'll revisit it tomorrow.

lib/fastly_nsq/testing.rb
```diff
@@ -278,6 +282,12 @@ def connected?
       @connected
     end
 
+    def connect(*args)
+      return super(*args) unless FastlyNsq::Testing.enabled?
+      @connected = true
+      FakeConnection.new
```
lanej: Probably want this to take the value of `@connected` right?

PR #94, opened 2018-10-04
Add support for configuring max_threads

leklund:
Reason for Change
===================
* `MAX_THREADS` is a constant and not easily configured. This allows applications to set the max number of threads for the thread pool easily.

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq/manager.rb
+++ b/lib/fastly_nsq/manager.rb
@@ -18,6 +18,7 @@ class FastlyNsq::Manager
   # Create a FastlyNsq::Manager
   #
   # @param logger [Logger]
+  # @param max_threads [Integer] Maxiumum number of threads to be used by {FastlyNsq::PriorityThreadPool}
   # @param pool_options [Hash] Options forwarded to {FastlyNsq::PriorityThreadPool} constructor.
   def initialize(logger: FastlyNsq.logger, max_threads: FastlyNsq.max_processing_pool_threads, **pool_options)
     @done      = false
--- a/lib/fastly_nsq/manager.rb
+++ b/lib/fastly_nsq/manager.rb
@@ -24,7 +24,7 @@ class FastlyNsq::Manager
     @done      = false
     @logger    = logger
     @pool      = FastlyNsq::PriorityThreadPool.new(
-      { fallback_policy: :caller_runs, max_threads: DEFAULT_POOL_SIZE }.merge(pool_options),
+      { fallback_policy: :caller_runs, max_threads: FastlyNsq.max_threads || DEFAULT_POOL_SIZE }.merge(pool_options),
     )
   end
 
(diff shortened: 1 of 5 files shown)


set5think (approved):

lib/fastly_nsq/manager.rb
```diff
@@ -24,7 +24,7 @@ def initialize(logger: FastlyNsq.logger, **pool_options)
     @done      = false
     @logger    = logger
     @pool      = FastlyNsq::PriorityThreadPool.new(
-      { fallback_policy: :caller_runs, max_threads: DEFAULT_POOL_SIZE }.merge(pool_options),
+      { fallback_policy: :caller_runs, max_threads: FastlyNsq.max_threads || DEFAULT_POOL_SIZE }.merge(pool_options),
```
lanej: I think maybe a little cleaner if you made this a named parameter in the method signature and defaulted `FastlyNsq.max_threads` to `DEFAULT_POOL_SIZE`

leklund: a named parameter for which method?

lanej: `#initialize`

leklund: oh. I didn't want to mess with that since it's just doing

```
def initialize(*)
  super
  
  @queue = FastlyNsq::PriorityQueue.new(:max)
end
```

lanej: oh sorry, I meant `FastlyNsq::Manager#initialize(logger: FastlyNsq.logger, **pool_options)` so

```ruby
class FastlyNsq::Manager
  def initialize(logger: FastlyNsq.logger, max_threads: FastlyNsq.max_threads, **pool_options)
	...
  end
end
```

leklund: ohhh. that makes a lot more sense

leklund:
@lanej updated the manager constuctor and also decided to rename the setting to `max_processing_pool_threads`. It's verbose but it is more descriptive/accurate than `max_threads`.

lib/fastly_nsq/manager.rb
```diff
@@ -20,11 +19,11 @@ class FastlyNsq::Manager
   #
   # @param logger [Logger]
   # @param pool_options [Hash] Options forwarded to {FastlyNsq::PriorityThreadPool} constructor.
-  def initialize(logger: FastlyNsq.logger, **pool_options)
+  def initialize(logger: FastlyNsq.logger, max_threads: FastlyNsq.max_processing_pool_threads, **pool_options)
```
DuoSRX, line 23: We should add `max_threads` to the doc string like `logger`.

leklund: whoops, forgot to do that. good catch

PR #95, opened 2019-08-15
Log the internal NSQ id

leklund:
Reason for Change
===================

WHen dealing with `E_FIN_FAILED` it may be useful to see when the oringal message was recieved.

List of Changes
===============
* delegate is to nsq_message
* log the id as nsq id

### DIFF
--- a/lib/fastly_nsq/listener.rb
+++ b/lib/fastly_nsq/listener.rb
@@ -116,8 +116,9 @@ class FastlyNsq::Listener
     msg_info = {
       channel:  channel,
       topic:    topic,
-      attempts: nsq_message.attempts,
+      attempts: message.attempts,
       id:       Digest::MD5.hexdigest(nsq_message.body.to_s),
+      nsq_id:   message.id,
       metadata: message.meta,
     }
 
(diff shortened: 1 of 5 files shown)


lib/fastly_nsq/listener.rb
```diff
@@ -116,8 +116,9 @@ def call(nsq_message)
     msg_info = {
       channel:  channel,
       topic:    topic,
-      attempts: nsq_message.attempts,
+      attempts: message.attempts,
```
alieander, line 119: Message and nsq_message are the same then?

alieander (approved):

lib/fastly_nsq/listener.rb
leklund: we delegate attempts in `FastlyNsq::Message` to the nsq_mesage. Figured I'd just use that interface

DuoSRX (approved):

lib/fastly_nsq/message.rb
```diff
@@ -20,6 +20,10 @@ class FastlyNsq::Message
   #   Delegated to +self.nsq_message+
   #   @return [Nsq::Message#timestamp]
   #   @see https://www.rubydoc.info/gems/nsq-ruby/Nsq/Message#timestamp-instance_method
+  # @!method id
+  #   Delegated to +self.nsq_message+
+  #   @return [Nsq::Message#id]
+  #   @see https://www.rubydoc.info/gems/nsq-ruby/Nsq/Message#id-instance_method
```
leklund, line 26: lol I updated the documentation but didn't add the *&^%ing delegator.

PR #98, opened 2021-01-27
Update priority queue to be a :min queue by default and add setting

leklund:
Summary
=======

Adds a new setting priority_queue_type that can be `:min` or `:max`. Update the default to be :min.

This is a breaking change for workflows that depend on a `:max` priority queue.

Recommended Reviewers
=====================
@fastly/billing

### DIFF
--- a/lib/fastly_nsq.rb
+++ b/lib/fastly_nsq.rb
@@ -123,6 +123,23 @@ module FastlyNsq
     end
 
     ##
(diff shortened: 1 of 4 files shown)


alieander (approved):

leklund:
Going to solve this by changing priorities instead of breaking the gem.

lanej:
👋🏻

PR #105, opened 2022-03-04
Configurable NSQDs for Consumers and Producers

leklund:
Reason for Change
===================
Sometimes you don't want to use nsqlookupd. This PR adds support for specifying a set of NSQD instances for Producers and Consumers.

This can be configured either via environment variables:

```
ENV['NSQD_CONSUMERS']="host1:4150,host2:4150"
ENV['NSQD_PRODUCERS']="host2:4150,host4:4150"
```

Or via `configure`:

```
  config.consumer_nsqds = ["nsqd01:4150", "nsqd02:4150"]
  config.producer_nsqds = ["nsqd01:4150", "nsqd02:4150"]
end
```

If, for example, you wanted get consumer nsqds from the lookups but specify a subset of nsqds for production you could do:
```
  config.lookupd_http_addresses = ["lookupd01:4161", "lookupd02:4161"]
  config.producer_nsqds = ["nsqd03:4150", "nsqd04:4150"]
end
```

List of Changes
===============
- add FastlyNsq.consumer_nsqds and FastlyNsq.producer_nsqds
- don't require the NSQD addresses to be set
- test cleanup -- fixes extra logging and rspec warnings
- add support for producer_nsqds to the FastlyNsq::Producer class
- add support for consumer_nsqds to the FastlyNsq::Consumer class
- version++

Risks
=====
* Low. This adds new config variables but will not be used unless they are set.


lib/fastly_nsq/consumer.rb
```diff
@@ -99,15 +102,25 @@ def empty?
...
+
+    opts = {
+      topic: topic,
+      channel: channel,
+      queue: queue,
+      max_attempts: max_attempts,
+      **options
+    }.merge(tls_options)
+
+    if !consumers.empty?
```
alieander, line 116: I assume if ! over unless because there is not a else unless ;)

spec/consumer_spec.rb
```diff
@@ -56,4 +56,45 @@
       it { should be_empty }
     end
   end
+
+  describe "connection priorioty" do
```
alieander: typo

alieander (approved):

lib/fastly_nsq/consumer.rb
leklund: `unless else` makes my brain hurt! 🤣

alieander (approved):

